PR #399, opened 2026-04-03
feat(chart): add Helm chart for kumo

kahirokunn:
## Summary
- add a Helm chart for deploying kumo

## Manual Test
Screenshot of `make test-helm-e2e` running.

CleanShot 2026-04-03 at 22 01 44@2x

Release looks good so far in [kahirokunn/kumo#3](https://github.com/kahirokunn/kumo/pull/3).

PR #400, opened 2026-04-06
feat(sqs): add ListQueueTags, TagQueue, and UntagQueue actions

kahirokunn:
Cherry-picked from #399 to narrow the scope of the Helm chart PR.


internal/service/sqs/storage.go
```diff
@@ -187,14 +191,27 @@ func (s *MemoryStorage) resolveQueueData(queueURL string) (string, *QueueData, e
 	return "", nil, ErrQueueDoesNotExist
 }
 
+func cloneStringMap(src map[string]string) map[string]string {
```
sivchari: Why do you need this ? You can just use maps.Clone directly.

PR #430, opened 2026-04-20
feat(route53): add GetChange API

apenney:
Disclaimer: All this code, the PR description, everything was churned out by AI (Opus 4.7).  I just needed the /2013-04-01/change/{id} endpoint to work, and stumbled my way via the power of delegation to a robot to coming up with something that seems to work.  I would be very happy to change or reimplement anything, I just didn't want to open a bug report and not at least try to make it work myself.

Adds the GET /2013-04-01/change/{id} (GetChange) endpoint to the Route 53 service emulator. AWS clients call this endpoint after ChangeResourceRecordSets, CreateHostedZone, and DeleteHostedZone to poll for change propagation status; previously kumo had no route registered and returned 404, which broke any client that followed the documented call pattern.

Storage:

- Add a Changes map on MemoryStorage keyed by the bare change ID (UUID without the "/change/" prefix) and persisted alongside the existing HostedZones / RecordSets via the existing JSON marshal/unmarshal hooks.
- Add an ErrChangeNotFound sentinel and PutChange / GetChange methods on both the Storage interface and MemoryStorage.
- Both methods strip a leading "/change/" so the handler is robust regardless of whether the AWS SDK's URL encoding preserves the prefix on the request path.

Handlers:

- Lift the previously inline ChangeInfo literals in CreateHostedZone, DeleteHostedZone, and ChangeResourceRecordSets into a local variable and persist via storage.PutChange before writing the response, so any change ID kumo emits is subsequently retrievable.
- Add the GetChange handler mirroring GetHostedZone: returns NoSuchChange (404) for unknown IDs, InvalidInput (400) for an empty path value, and InternalError on unexpected storage errors.

Types:

- Add GetChangeResponse matching the AWS XML shape: <GetChangeResponse xmlns="https://route53.amazonaws.com/doc/2013-04-01/"> <ChangeInfo><Id/><Status/><SubmittedAt/><Comment/></ChangeInfo> </GetChangeResponse> reusing the existing ChangeInfo type.

Routing:

- Register GET /2013-04-01/change/{id} in RegisterRoutes.

Tests:

- TestRoute53_GetChange: creates a hosted zone, then calls GetChange with the change ID returned in the create response and asserts the response payload via a golden fixture (Id and SubmittedAt ignored since both are non-deterministic).
- TestRoute53_GetChange_NotFound: asserts the AWS SDK surfaces a *types.NoSuchChange error for unknown change IDs.

Status is always returned as "INSYNC", matching the existing behavior of the three handlers that emit a ChangeInfo. Simulating the brief PENDING -> INSYNC transition that real AWS exhibits is out of scope for this change.

### DIFF
--- a/internal/service/route53/storage.go
+++ b/internal/service/route53/storage.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 	"sync"
 
 	"github.com/sivchari/kumo/internal/storage"
@@ -22,6 +23,8 @@ var (
 	ErrRecordSetAlreadyExists = errors.New("record set already exists")
 	// ErrInvalidInput is returned when input is invalid.
 	ErrInvalidInput = errors.New("invalid input")
+	// ErrChangeNotFound is returned when a change is not found.
+	ErrChangeNotFound = errors.New("change not found")
 )
 
 // Storage defines the interface for Route 53 storage operations.
@@ -32,6 +35,8 @@ type Storage interface {
 	DeleteHostedZone(id string) error
 	GetRecordSets(hostedZoneID string) ([]ResourceRecordSet, error)
 	ChangeRecordSets(hostedZoneID string, changes []Change) error
+	PutChange(change *ChangeInfo) error
+	GetChange(id string) (*ChangeInfo, error)
 }
 
 // Option is a configuration option for MemoryStorage.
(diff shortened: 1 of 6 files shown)


internal/service/route53/storage.go
```diff
@@ -257,6 +268,32 @@ func (s *MemoryStorage) ChangeRecordSets(hostedZoneID string, changes []Change)
...
+	defer s.mu.Unlock()
+
+	id := strings.TrimPrefix(change.ID, "/change/")
+	s.Changes[id] = change
+
+	return nil
+}
+
+// GetChange retrieves a change by ID.
+func (s *MemoryStorage) GetChange(id string) (*ChangeInfo, error) {
```
sivchari, line 283: The handler passes path value, so strings.TrimPrefix is not necessary.

apenney: Pushed a commit on this, thanks!

sivchari (approved):
Thx!

PR #509, opened 2026-05-04
feat: CodeCommit service implementation and integration tests

### LINKED ISSUE #52: CodeCommit: Service implementation
## Overview
Implement CodeCommit service emulator.

## Target APIs
- `CreateRepository`
- `DeleteRepository`
- `GetRepository`
- `ListRepositories`
- `CreateBranch`
- `GetBranch`
- `ListBranches`
- `GetFile`
- `PutFile`

## References
- https://docs.aws.amazon.com/codecommit/latest/APIReference/

matheusandre1:
Closes: #52

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -89,6 +89,10 @@ linters:
       - path: internal/service/codeconnections/types\.go
         linters:
           - tagliatelle
+      # AWS CodeCommit API requires PascalCase JSON field names.
+      - path: internal/service/codecommit/types\.go
+        linters:
+          - tagliatelle
       # AWS KMS API requires PascalCase JSON field names.
       - path: internal/service/kms/types\.go
         linters:
--- a/go.mod
+++ b/go.mod
@@ -3,46 +3,45 @@ module github.com/sivchari/kumo
 go 1.25.0
 
 require (
+	github.com/aws/aws-sdk-go-v2 v1.41.7
+	github.com/aws/aws-sdk-go-v2/config v1.32.17
+	github.com/aws/aws-sdk-go-v2/credentials v1.19.16
+	github.com/aws/aws-sdk-go-v2/service/acm v1.38.3
+	github.com/aws/aws-sdk-go-v2/service/amplify v1.38.16
+	github.com/aws/aws-sdk-go-v2/service/apigateway v1.39.3
+	github.com/aws/aws-sdk-go-v2/service/appmesh v1.35.14
+	github.com/aws/aws-sdk-go-v2/service/appsync v1.53.7
+	github.com/aws/aws-sdk-go-v2/service/athena v1.57.6
+	github.com/aws/aws-sdk-go-v2/service/backup v1.55.2
+	github.com/aws/aws-sdk-go-v2/service/dynamodb v1.57.3
+	github.com/aws/aws-sdk-go-v2/service/eventbridge v1.45.25
+	github.com/aws/aws-sdk-go-v2/service/kinesis v1.43.7
+	github.com/aws/aws-sdk-go-v2/service/kms v1.51.1
+	github.com/aws/aws-sdk-go-v2/service/s3 v1.100.1
+	github.com/aws/aws-sdk-go-v2/service/sqs v1.42.27
 	github.com/fxamacker/cbor/v2 v2.9.0
 	github.com/google/uuid v1.6.0
+	github.com/spf13/cobra v1.10.2
 	golang.org/x/crypto v0.47.0
 )
 
 require (
-	github.com/aws/aws-sdk-go-v2 v1.41.7 // indirect
 	github.com/aws/aws-sdk-go-v2/aws/protocol/eventstream v1.7.10 // indirect
-	github.com/aws/aws-sdk-go-v2/config v1.32.17 // indirect
-	github.com/aws/aws-sdk-go-v2/credentials v1.19.16 // indirect
 	github.com/aws/aws-sdk-go-v2/feature/ec2/imds v1.18.23 // indirect
 	github.com/aws/aws-sdk-go-v2/internal/configsources v1.4.23 // indirect
 	github.com/aws/aws-sdk-go-v2/internal/endpoints/v2 v2.7.23 // indirect
 	github.com/aws/aws-sdk-go-v2/internal/v4a v1.4.24 // indirect
-	github.com/aws/aws-sdk-go-v2/service/acm v1.38.3 // indirect
-	github.com/aws/aws-sdk-go-v2/service/amplify v1.38.16 // indirect
-	github.com/aws/aws-sdk-go-v2/service/apigateway v1.39.3 // indirect
-	github.com/aws/aws-sdk-go-v2/service/appmesh v1.35.14 // indirect
-	github.com/aws/aws-sdk-go-v2/service/appsync v1.53.7 // indirect
-	github.com/aws/aws-sdk-go-v2/service/athena v1.57.6 // indirect
-	github.com/aws/aws-sdk-go-v2/service/backup v1.55.2 // indirect
-	github.com/aws/aws-sdk-go-v2/service/dynamodb v1.57.3 // indirect
-	github.com/aws/aws-sdk-go-v2/service/eventbridge v1.45.25 // indirect
 	github.com/aws/aws-sdk-go-v2/service/internal/accept-encoding v1.13.9 // indirect
 	github.com/aws/aws-sdk-go-v2/service/internal/checksum v1.9.15 // indirect
 	github.com/aws/aws-sdk-go-v2/service/internal/endpoint-discovery v1.11.23 // indirect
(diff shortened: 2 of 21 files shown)


internal/service/codecommit/handlers.go
```diff
@@ -0,0 +1,451 @@
...
+		commitID = commit.CommitID
+	}
+
+	writeJSONResponse(w, GetFileResponse{
+		CommitID:    commitID,
+		BlobID:      uuid.New().String(),
+		FilePath:    file.FilePath,
+		FileMode:    file.FileMode,
+		FileSize:    int64(len(file.FileContent)),
+		FileContent: base64.StdEncoding.EncodeToString(file.FileContent),
```
sivchari, line 243: The AWS SDK automatically base64-encodes []byte fields when serializing to JSON, so PutFileRequest.FileContent already contains a base64-encoded string by the time it reaches the server. The current implementation stores this string as-is via []byte(fileContent), and then GetFile applies base64.StdEncoding.EncodeToString() again, resulting in double encoding.

The golden file confirms this — "FileContent": "U0dWc2JHOHNJRmR2Y214a0lRPT0=" decodes to "SGVsbG8sIFdvcmxkIQ==", which further decodes to "Hello, World!". The FileSize of 20 also reflects the length of the base64 string rather than the actual data size of 13.

To fix this, the PutFile handler should decode the base64 string before passing it to storage, while the GetFile handler can remain as-is. This matches the existing pattern used in the Kinesis service.

internal/service/codecommit/handlers.go
```diff
@@ -0,0 +1,451 @@
...
+	w.Header().Set("x-amzn-RequestId", uuid.New().String())
+	w.WriteHeader(status)
+	_ = json.NewEncoder(w).Encode(ErrorResponse{
+		Type:    code,
+		Message: message,
+	})
+}
+
+// handleCodeCommitError handles a CodeCommit error and writes the response.
+func handleCodeCommitError(w http.ResponseWriter, err error) {
```
sivchari, line 348: According to the [AWS documentation](https://docs.aws.amazon.com/codecommit/latest/APIReference/API_CreateRepository.html), EncryptionIntegrityChecksFailedException is the only error that returns HTTP 500. All other errors return 400. The current switch has several cases that all set status = http.StatusBadRequest, which is already the default value, making them redundant. It would be cleaner to only handle the 500 cases explicitly and let the default cover 400.

Several error code constants like errRepositoryNameRequiredException and errInvalidFileLocationException are defined in storage.go but never returned from any storage method. For example, handlers currently call writeCodeCommitError directly with string literals like errInvalidRepositoryNameException for validation, but the corresponding constants in storage.go are left unused. It would be more consistent to have the storage layer return ServiceError with these codes so that all error handling flows through handleCodeCommitError.

test/integration/codecommit_test.go
```diff
@@ -0,0 +1,383 @@
...
+	_, err = client.GetFile(ctx, &codecommit.GetFileInput{
+		RepositoryName: aws.String("test-filenotfound-repo"),
+		FilePath:       aws.String("non-existent.txt"),
+	})
+	if err == nil {
+		t.Fatal("expected error for non-existent file")
+	}
+}
+
+func TestCodeCommit_PutFileWithTags(t *testing.T) {
```
sivchari: This test is actually testing CreateRepository with tags rather than PutFile, so renaming it to TestCodeCommit_CreateRepositoryWithTags would be more accurate.

internal/service/codecommit/storage.go
```diff
@@ -0,0 +1,589 @@
...
+	if _, exists := branches[branchName]; exists {
+		return nil, &ServiceError{
+			Code:    errBranchNameExistsException,
+			Message: fmt.Sprintf("A branch named %s already exists.", branchName),
+		}
+	}
+
+	if commitID == "" {
+		commitID = branches["main"].DefaultCommitID
+	}
```
sivchari, line 379: Per the [AWS docs](https://docs.aws.amazon.com/codecommit/latest/APIReference/API_CreateBranch.html), `commitId` is **Required: Yes** for CreateBranch. When it's missing or empty, AWS returns `CommitIdRequiredException` (HTTP 400). This fallback to main's commit diverges from the spec.

Please return `CommitRequiredException` (or add `CommitIdRequiredException`) when `commitID` is empty, instead of silently falling back.

test/integration/codecommit_test.go
```diff
@@ -0,0 +1,383 @@
...
+	t.Cleanup(func() {
+		_, _ = client.DeleteRepository(context.Background(), &codecommit.DeleteRepositoryInput{
+			RepositoryName: aws.String("test-branch-repo"),
+		})
+	})
+
+	_, err = client.CreateBranch(ctx, &codecommit.CreateBranchInput{
+		RepositoryName: aws.String("test-branch-repo"),
+		BranchName:     aws.String("feature-branch"),
+		CommitId:       aws.String(""),
```
sivchari: `CommitId: aws.String("")` would return `CommitIdRequiredException` on real AWS. This test should first retrieve the actual commit ID (e.g., via `GetBranch` for "main") and pass it here.

internal/service/codecommit/storage.go
```diff
@@ -0,0 +1,589 @@
...
+	errInvalidRepositoryNameException           = "InvalidRepositoryNameException"
+	errBranchDoesNotExistException              = "BranchDoesNotExistException"
+	errBranchNameExistsException                = "BranchNameExistsException"
+	errBranchNameRequiredException              = "BranchNameRequiredException"
+	errInvalidBranchNameException               = "InvalidBranchNameException"
+	errFileDoesNotExistException                = "FileDoesNotExistException"
+	errFileContentRequiredException             = "FileContentRequiredException"
+	errInvalidFileLocationException             = "InvalidFileLocationException"
+	errInvalidFilePathException                 = "InvalidFilePathException"
+	errSamePathRequestException                 = "SamePathRequestException"
```
sivchari: Per the AWS docs for the implemented APIs:

- `errSamePathRequestException` (line 31): Not documented in any of the 9 implemented APIs. It's for merge operations only. Remove.
- `errInvalidFileLocationException` (line 29): Same — merge-only error. Remove.
- `errInvalidFilePathException` (line 30): Not the correct error code name. AWS uses [`InvalidPathException`](https://docs.aws.amazon.com/codecommit/latest/APIReference/API_GetFile.html) for GetFile/PutFile. Rename to the correct name.

Conversely, `errInvalidRepositoryNameException` (line 22) is documented in almost all implemented APIs but is never used. Please add repository name validation (e.g., regex `[\w\.-]+`, max 100 chars, `.git` suffix prohibited) and return this error.

.golangci.yaml
```diff
@@ -89,6 +89,10 @@ linters:
       - path: internal/service/codeconnections/types\.go
         linters:
           - tagliatelle
+      # AWS CodeCommit API requires PascalCase JSON field names.
```
sivchari, line 92: This comment is inaccurate. CodeCommit actually uses camelCase for most fields; only `Arn` is PascalCase (confirmed in the [official RepositoryMetadata spec](https://docs.aws.amazon.com/codecommit/latest/APIReference/API_RepositoryMetadata.html)).

Suggest: `# AWS CodeCommit API uses mixed-case JSON field names (Arn is PascalCase).`

internal/service/codecommit/storage.go
```diff
@@ -0,0 +1,589 @@
...
+
+	return nil
+}
+
+// CreateRepository creates a new repository.
+func (s *MemoryStorage) CreateRepository(_ context.Context, name, description string, tags map[string]string, kmsKeyID string) (*Repository, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if name == "" {
```
sivchari, line 169: AWS returns `InvalidRepositoryNameException` when the repository name doesn't match `[\w\.-]+`, exceeds 100 characters, or ends with `.git`. Currently only empty name is checked (`RepositoryNameRequiredException`), but invalid names are not validated.

Ref: https://docs.aws.amazon.com/codecommit/latest/APIReference/API_CreateRepository.html

go.mod
```diff
@@ -42,6 +43,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/sts v1.42.1 // indirect
 	github.com/aws/smithy-go v1.25.1 // indirect
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
+	github.com/sivchari/golden v0.3.0 // indirect
```
sivchari: `golden` is a test-only dependency used in `test/go.mod`. Is it needed in the root `go.mod`?

sivchari (changes requested):
Thanks for the implementation! I verified the behavior against the [AWS official documentation](https://docs.aws.amazon.com/codecommit/latest/APIReference/Welcome.html) and found some deviations from the spec. Please see the inline comments below.

Issue #408, opened 2026-04-07
IAM: Add group, instance profile, and policy version operations


sivchari:
## Description
Add missing IAM operations for groups, instance profiles, and policy version management.

### High Priority
- [ ] CreateGroup / DeleteGroup / GetGroup / ListGroups
- [ ] AddUserToGroup / RemoveUserFromGroup / ListGroupsForUser
- [ ] CreateInstanceProfile / DeleteInstanceProfile / GetInstanceProfile / ListInstanceProfiles
- [ ] AddRoleToInstanceProfile / RemoveRoleFromInstanceProfile

### Medium Priority
- [ ] CreatePolicyVersion / GetPolicyVersion / DeletePolicyVersion / ListPolicyVersions / SetDefaultPolicyVersion
- [ ] PutUserPolicy / GetUserPolicy / DeleteUserPolicy / ListUserPolicies
- [ ] PutRolePolicy / GetRolePolicy / DeleteRolePolicy / ListRolePolicies
- [ ] UpdateAssumeRolePolicy
- [ ] ListAttachedUserPolicies / ListAttachedRolePolicies / ListAttachedGroupPolicies

### Low Priority
- [ ] AttachGroupPolicy / DetachGroupPolicy / PutGroupPolicy / GetGroupPolicy / DeleteGroupPolicy
- [ ] CreateLoginProfile / GetLoginProfile / UpdateLoginProfile / DeleteLoginProfile
- [ ] UpdateUser / UpdateRole
- [ ] GetAccessKeyLastUsed / UpdateAccessKey
- [ ] TagUser / UntagUser / TagRole / UntagRole / ListUserTags / ListRoleTags

## Notes
Currently 18 operations are implemented. Groups and instance profiles are essential for realistic IAM testing.

PR #551, opened 2026-05-08
feat(iam): ListInstanceProfilesForRole stub for role tear-down

mizchi:
## Summary

Adds an empty-list \`ListInstanceProfilesForRole\` handler. AWS clients call this during role deletion to detach any instance profiles before issuing \`DeleteRole\`. kumo does not model instance profiles, so the correct emulator response is an empty list — the same response real IAM gives for a role that never had a profile.

Without this, a role created via kumo cannot be cleanly deleted by clients that follow the standard tear-down idiom: the list call returns \`InvalidAction\` and the client aborts before reaching \`DeleteRole\`.

## Implementation

- One handler (\`ListInstanceProfilesForRole\`) with the AWS-spec XML response shape (\`<InstanceProfiles>\` with no members, \`<IsTruncated>false</IsTruncated>\`).
- One register entry in \`initActionHandlers\`.
- No storage changes — instance profiles are still not modeled, by design.

## Test plan

- [x] \`make build\`
- [x] All existing \`TestIAM_*\` integration tests still pass
- [x] New \`TestIAM_ListInstanceProfilesForRole\` creates a role, calls the new action, and asserts the returned list is empty

---

Related to sivchari/kumo#408 — High Priority "CreateInstanceProfile / DeleteInstanceProfile / GetInstanceProfile / ListInstanceProfiles" / "AddRoleToInstanceProfile / RemoveRoleFromInstanceProfile". This PR only adds the read-side stub used during role tear-down. The full instance-profile lifecycle is a separate follow-up.

### DIFF
--- a/internal/service/iam/handlers.go
+++ b/internal/service/iam/handlers.go
@@ -485,6 +485,22 @@ func (s *Service) ListAccessKeys(w http.ResponseWriter, r *http.Request) {
 	})
 }
 
+// ListInstanceProfilesForRole returns an empty list. Instance profiles are
+// not modeled; AWS clients call this during role tear-down to detach profiles
+// before deletion, so an empty list is the correct "no profiles" response.
+func (s *Service) ListInstanceProfilesForRole(w http.ResponseWriter, _ *http.Request) {
+	writeIAMXMLResponse(w, struct {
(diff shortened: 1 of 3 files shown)


sivchari (approved):

sivchari:
Superseded by #555 which includes the full implementation.

PR #553, opened 2026-05-08
feat(ec2): tags-on-create via TagSpecifications on Create* actions

mizchi:
## Summary

The AWS EC2 SDK accepts a \`TagSpecifications\` parameter on the \`Create*\` actions so callers can tag a resource in the same call that creates it. Adds parsing + forwarding for this pattern on the five Create handlers that currently model taggable resources.

| Action | Resource type accepted in \`TagSpecifications\` |
|---|---|
| \`CreateVpc\` | \`vpc\` |
| \`CreateSubnet\` | \`subnet\` |
| \`CreateInternetGateway\` | \`internet-gateway\` |
| \`CreateRouteTable\` | \`route-table\` |
| \`CreateSecurityGroup\` | \`security-group\` |

Without this, tags from the create call were silently dropped: a subsequent \`Describe*\` showed an empty \`TagSet\`, which downstream parity checks (re-applying the same config, or comparing CloudFormation drift) flag as a difference.

## Implementation

- \`applyTagsOnCreate\` helper called after each \`Create*\`: parses \`TagSpecifications\` from the form and forwards to \`storage.CreateTags\`. Since the existing \`storage.Create*\` methods return the storage-owned pointer, and \`storage.CreateTags\` writes through that pointer, the response struct reflects the tags by the time the handler builds the XML response — no separate response-mutation step needed.
- \`parseTagSpecificationsForResourceType\` reads the \`TagSpecifications.N.ResourceType\` / \`.Tag.M.{Key,Value}\` pattern directly from \`r.Form\`. The Query form-to-JSON converter does not understand this nested pattern, so this mirrors the approach already used by \`CreateTags\` / \`DescribeTags\` in #547.

## Stacking

This branch is based on \`feat/ec2-tag-apis\` (PR #547) because it calls \`storage.CreateTags\` introduced there. If #547 lands first, this rebases cleanly.

## Test plan

- [x] \`make build\`
- [x] All existing \`TestEC2_*\` integration tests still pass
- [x] New \`TestEC2_CreateVpc_WithTagSpecifications\` exercises the full path: tags supplied to \`CreateVpc\` are reflected in the create response, in \`DescribeVpcs\`, and in \`DescribeTags\` filtered by \`resource-id\`

### DIFF
--- a/internal/service/ec2/handlers.go
+++ b/internal/service/ec2/handlers.go
@@ -213,6 +213,8 @@ func (s *Service) CreateSecurityGroup(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	applyTagsOnCreate(r, s.storage, sg.GroupID, "security-group")
+
 	writeEC2XMLResponse(w, XMLCreateSecurityGroupResponse{
 		Xmlns:     ec2XMLNS,
 		RequestID: uuid.New().String(),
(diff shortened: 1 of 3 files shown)


sivchari (approved):

sivchari:
Like the others.

sivchari (approved):

Issue #409, opened 2026-04-07
EC2: Add security group describe, tags, volumes, and snapshots


sivchari:
## Description
Add missing EC2 operations for security group queries, tagging, volumes, and snapshots.

### High Priority
- [ ] DescribeSecurityGroups / DescribeSecurityGroupRules
- [ ] RevokeSecurityGroupIngress / RevokeSecurityGroupEgress
- [ ] CreateTags / DeleteTags / DescribeTags
- [ ] DescribeAvailabilityZones
- [ ] DescribeImages

### Medium Priority
- [ ] CreateVolume / DeleteVolume / DescribeVolumes / AttachVolume / DetachVolume
- [ ] CreateSnapshot / DeleteSnapshot / DescribeSnapshots
- [ ] CreateNetworkInterface / DeleteNetworkInterface / DescribeNetworkInterfaces
- [ ] ModifyInstanceAttribute
- [ ] RebootInstances
- [ ] DetachInternetGateway / DeleteInternetGateway
- [ ] DeleteNatGateway

### Low Priority
- [ ] DescribeRegions
- [ ] ImportKeyPair
- [ ] AllocateAddress / ReleaseAddress / AssociateAddress / DisassociateAddress / DescribeAddresses
- [ ] ModifyVpcAttribute / DescribeVpcAttribute
- [ ] ModifySubnetAttribute

## Notes
Currently 34 operations are implemented. DescribeSecurityGroups and CreateTags/DescribeTags are among the most commonly used EC2 operations missing.

PR #554, opened 2026-05-08
feat(ec2): VPC/Subnet attribute mutation (Modify/Describe)

mizchi:
## Summary

Adds the attribute APIs the AWS EC2 SDK uses to read and toggle per-VPC and per-Subnet flags:

| Action | Attribute(s) |
|---|---|
| \`ModifyVpcAttribute\` | \`EnableDnsHostnames\`, \`EnableDnsSupport\` |
| \`DescribeVpcAttribute\` | same plus \`enableNetworkAddressUsageMetrics\` (returns \`false\` — not modeled) |
| \`ModifySubnetAttribute\` | \`MapPublicIpOnLaunch\` (\`AssignIpv6AddressOnCreation\` accepted but not modeled) |

These are required for any client that wants to confirm or change the \`EnableDnsHostnames\` flag on a freshly-created VPC, or set \`MapPublicIpOnLaunch\` on a subnet that should auto-assign public IPs.

## Behavior alignment with AWS

- The \`Vpc\` struct gains \`EnableDNSHostnames\` and \`EnableDNSSupport\` fields. \`CreateVpc\` defaults \`EnableDNSSupport=true\` to match real AWS — without this, clients that call \`DescribeVpcAttribute\` immediately after \`CreateVpc\` see a VPC that drifts from a real one (\`enable_dns_support\` reads back \`false\` instead of \`true\`).
- AWS modifies one attribute per call. \`VpcAttributeUpdates\` / \`SubnetAttributeUpdates\` use \`*bool\` pointers so storage applies only what the request supplied.
- Form keys (\`EnableDnsHostnames.Value\` etc.) follow the AWS wire format. The Go struct fields use \`EnableDNS*\` per Go naming conventions.
- \`enableNetworkAddressUsageMetrics\` is queried by recent SDKs even on plain VPCs; this returns \`false\` rather than an error, mirroring a non-NAUM-enabled real VPC.

## Test plan

- [x] \`make build\`
- [x] All existing \`TestEC2_*\` integration tests still pass
- [x] New \`TestEC2_ModifyAndDescribeVpcAttribute\` asserts the AWS-default \`EnableDnsSupport=true\` / \`EnableDnsHostnames=false\` on \`CreateVpc\`, then toggles \`EnableDnsHostnames\` and reads it back
- [x] New \`TestEC2_ModifySubnetAttribute\` toggles \`MapPublicIpOnLaunch\` and reads it back via \`DescribeSubnets\`

---

Addresses sivchari/kumo#409 — Low Priority "ModifyVpcAttribute / DescribeVpcAttribute" and "ModifySubnetAttribute".

### DIFF
--- a/internal/service/ec2/handlers.go
+++ b/internal/service/ec2/handlers.go
@@ -966,6 +966,178 @@ func (s *Service) DescribeTags(w http.ResponseWriter, r *http.Request) {
 	})
 }
 
+// ModifyVpcAttribute handles the ModifyVpcAttribute action. AWS modifies one
+// attribute per call; the handler updates only the fields the request supplies.
+func (s *Service) ModifyVpcAttribute(w http.ResponseWriter, r *http.Request) {
+	if err := r.ParseForm(); err != nil {
+		writeError(w, errInvalidParameter, "Failed to parse form data", http.StatusBadRequest)
+
+		return
+	}
+
+	vpcID := r.Form.Get("VpcId")
+	if vpcID == "" {
+		writeError(w, errInvalidParameter, "VpcId is required", http.StatusBadRequest)
+
+		return
+	}
+
+	if err := s.storage.ModifyVpcAttribute(r.Context(), vpcID, vpcAttributeUpdates(r.Form)); err != nil {
+		handleError(w, err)
+
+		return
+	}
+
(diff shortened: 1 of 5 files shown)


sivchari (approved):

sivchari:
Like the others

PR #555, opened 2026-05-08
feat(iam): instance profile lifecycle (Create/Get/Delete/List + AddRole/RemoveRole)

mizchi:
## Summary

Adds the IAM instance profile lifecycle. AWS uses \`InstanceProfile\` as the addressable identity that an EC2 instance assumes a role through, so without it any client that creates a profile, attaches a role, and queries it back fails immediately with \`InvalidAction\`.

| Operation | Notes |
|---|---|
| \`CreateInstanceProfile\` | Generates ARN as \`arn:aws:iam::<account>:instance-profile<path><name>\`. |
| \`DeleteInstanceProfile\` | Returns \`DeleteConflict\` if a role is still attached, matching AWS. |
| \`GetInstanceProfile\` | Returns the full profile including its current \`Roles\` list. |
| \`ListInstanceProfiles\` | Supports the \`PathPrefix\` filter; pagination not modeled. |
| \`ListInstanceProfilesForRole\` | Real implementation — returns profiles that contain the role. (Supersedes the empty stub in #551.) |
| \`AddRoleToInstanceProfile\` | Rejects a second role with \`LimitExceeded\` — AWS allows at most one role per profile. |
| \`RemoveRoleFromInstanceProfile\` | Plain detach. |

## Storage

\`MemoryStorage\` gains an \`InstanceProfiles map[string]*InstanceProfile\` keyed by name. Roles are referenced inline in the \`Roles []Role\` field on \`InstanceProfile\` (matching the AWS API shape, not normalized).

## Relationship to #551

#551 adds a stand-alone empty-list stub of \`ListInstanceProfilesForRole\` so role tear-down works in environments that never created a profile. This PR replaces the stub with a real implementation. If #551 lands first, this PR resolves the conflict by overriding the handler. If this lands first, #551 becomes a no-op and can be closed.

## Test plan

- [x] \`make build\`
- [x] All existing \`TestIAM_*\` integration tests still pass
- [x] New \`TestIAM_InstanceProfileLifecycle\` exercises Create → AddRole → GetInstanceProfile (asserts the role appears) → ListInstanceProfilesForRole (asserts the profile appears) → RemoveRole → GetInstanceProfile (asserts empty Roles)

---

Addresses sivchari/kumo#408 — High Priority "CreateInstanceProfile / DeleteInstanceProfile / GetInstanceProfile / ListInstanceProfiles" and "AddRoleToInstanceProfile / RemoveRoleFromInstanceProfile". Also covers the Medium Priority "ListInstanceProfilesForRole" via the real implementation.

### DIFF
--- a/internal/service/iam/handlers.go
+++ b/internal/service/iam/handlers.go
@@ -485,6 +485,151 @@ func (s *Service) ListAccessKeys(w http.ResponseWriter, r *http.Request) {
 	})
 }
 
+// ListInstanceProfilesForRole returns the profiles attached to a role.
+func (s *Service) ListInstanceProfilesForRole(w http.ResponseWriter, r *http.Request) {
+	roleName := getFormValue(r, "RoleName")
+	if roleName == "" {
+		writeIAMError(w, errInvalidParameter, "RoleName is required", http.StatusBadRequest)
+
+		return
+	}
+
+	profiles, err := s.storage.ListInstanceProfilesForRoleReal(r.Context(), roleName)
+	if err != nil {
+		handleIAMError(w, err)
+
+		return
+	}
+
+	writeIAMXMLResponse(w, ListInstanceProfilesForRoleResponseV2{
+		ListInstanceProfilesForRoleResult: ListInstanceProfilesForRoleResultV2{InstanceProfiles: profiles},
+		ResponseMetadata:                  ResponseMetadata{RequestID: uuid.New().String()},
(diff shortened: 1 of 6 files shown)


sivchari (approved):

sivchari:
Since the merge order failed and caused a conflict, I resolved the conflict instead.

sivchari (approved):

PR #557, opened 2026-05-08
feat(elbv2): LoadBalancer / TargetGroup attribute APIs (Modify/Describe)

mizchi:
## Summary

Adds the attribute APIs that AWS clients call after \`CreateLoadBalancer\` and \`CreateTargetGroup\` to read or change per-resource configuration:

| Action | Notes |
|---|---|
| \`ModifyLoadBalancerAttributes\` | Upserts the supplied attributes; returns the merged set. |
| \`DescribeLoadBalancerAttributes\` | Returns all attributes; lazy-initializes to AWS defaults if Modify has never been called. |
| \`ModifyTargetGroupAttributes\` | Same shape for target groups. |
| \`DescribeTargetGroupAttributes\` | Same. |

Without these, any client that sets attributes (\`deletion_protection.enabled\`, \`idle_timeout.timeout_seconds\`, \`deregistration_delay.timeout_seconds\`, \`stickiness.*\`, etc.) hits \`InvalidAction\` immediately after \`Create*\`, and any client that reads state back for drift detection sees an empty attribute set instead of the AWS defaults.

## Behavior alignment with AWS

\`Describe*Attributes\` lazily seeds the AWS-default attribute set so a freshly-created LB / TG reads back the same defaults real AWS surfaces:

| Resource | Default attributes (excerpt) |
|---|---|
| Load balancer | \`deletion_protection.enabled=false\`, \`idle_timeout.timeout_seconds=60\`, \`routing.http2.enabled=true\`, \`load_balancing.cross_zone.enabled=true\` |
| Target group | \`deregistration_delay.timeout_seconds=300\`, \`stickiness.enabled=false\`, \`stickiness.type=lb_cookie\`, \`load_balancing.algorithm.type=round_robin\` |

\`Modify*Attributes\` upserts; subsequent \`Describe\` returns the merged set sorted by key for deterministic output.

## Implementation

\`Attributes.member.N.{Key,Value}\` parsing lives in \`handlers.go\` in the same form-helper style as the existing CreateTags / DescribeTags code (PR #547). No change to the generic Query dispatcher.

\`LoadBalancer\` and \`TargetGroup\` storage types gain \`Attributes map[string]string\`. Defaults are seeded lazily so existing tests that build these resources directly do not need updating.

## Test plan

- [x] \`make build\`
- [x] All existing \`TestELBv2_*\` integration tests still pass
- [x] New \`TestELBv2_LoadBalancerAttributes\` asserts AWS defaults on first Describe, then toggles two attributes and reads them back
- [x] New \`TestELBv2_TargetGroupAttributes\` does the same for the TG side

### DIFF
--- a/internal/service/elbv2/handlers.go
+++ b/internal/service/elbv2/handlers.go
@@ -822,21 +822,25 @@ func (s *Service) DispatchAction(w http.ResponseWriter, r *http.Request) {
 // getActionHandler returns the handler function for the given action.
 func (s *Service) getActionHandler(action string) func(http.ResponseWriter, *http.Request) {
 	handlers := map[string]func(http.ResponseWriter, *http.Request){
-		"CreateLoadBalancer":    s.CreateLoadBalancer,
-		"DeleteLoadBalancer":    s.DeleteLoadBalancer,
-		"DescribeLoadBalancers": s.DescribeLoadBalancers,
-		"CreateTargetGroup":     s.CreateTargetGroup,
-		"DeleteTargetGroup":     s.DeleteTargetGroup,
-		"DescribeTargetGroups":  s.DescribeTargetGroups,
-		"RegisterTargets":       s.RegisterTargets,
-		"DeregisterTargets":     s.DeregisterTargets,
-		"CreateListener":        s.CreateListener,
-		"DeleteListener":        s.DeleteListener,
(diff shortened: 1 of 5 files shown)


sivchari (approved):

sivchari:
Like the others

sivchari (approved):

PR #567, opened 2026-05-08
feat(elbv2): empty-payload stubs for read-only metadata APIs

mizchi:
## Summary

The AWS provider's \`aws_lb\` / \`aws_lb_listener\` / \`aws_lb_target_group\` resources call several metadata APIs on every refresh that kumo does not implement yet. Without them, a freshly-created resource fails its very first read with \`api error UnknownError: UnknownError\` (the SDK's catch-all for InvalidAction-shaped XML errors). In the capacity-reservation case, an under-modeled response surfaces as perpetual drift.

This adds minimal handlers that respond with the AWS-default empty shape:

- \`DescribeTags\` — echoes each requested \`ResourceArn\` back with an empty \`<Tags/>\` child
- \`AddTags\` / \`RemoveTags\` — accept the request and return an empty body
- \`DescribeCapacityReservation\` — returns \`DecreaseRequestsRemaining=0\` and an empty \`CapacityReservationState\`. \`MinimumLoadBalancerCapacity\` is omitted on purpose: emitting it with \`CapacityUnits=0\` triggers perpetual drift in the provider.
- \`DescribeListenerAttributes\` / \`ModifyListenerAttributes\` — empty \`Attributes\` member list (both share the same handler since the wire shapes are identical)

## Why

These are pure wire-level read-path fixes. Tag and capacity state are not modeled by kumo; per-attribute / per-tag behaviour can be added later without changing the wire shape.

Caught while running \`terraform-provider-aws\` (5.x) end-to-end against kumo. With the handlers in place, an ALB lifecycle (\`apply\` → \`plan\` shows no drift → \`destroy\`) completes against kumo without intervention.

## Test plan

- [x] \`go build ./...\` passes
- [x] Existing ELBv2 integration tests still pass
- [ ] Reviewer: confirm the empty-payload shapes match what the SDK expects (especially \`DescribeCapacityReservation\` — see comment about omitting \`MinimumLoadBalancerCapacity\`)

### DIFF
--- a/internal/service/elbv2/handlers.go
+++ b/internal/service/elbv2/handlers.go
@@ -844,9 +844,129 @@ func (s *Service) getActionHandler(action string) func(http.ResponseWriter, *htt
 		"DescribeListeners":              s.DescribeListeners,
 		"ModifyListener":                 s.ModifyListener,
 		"DescribeTargetHealth":           s.DescribeTargetHealth,
+		"DescribeTags":                   s.DescribeTags,
+		"AddTags":                        s.tagsNoOp,
+		"RemoveTags":                     s.tagsNoOp,
+		"DescribeCapacityReservation":    s.DescribeCapacityReservation,
+		"DescribeListenerAttributes":     s.DescribeListenerAttributes,
+		"ModifyListenerAttributes":       s.DescribeListenerAttributes,
 	}
 
 	return handlers[action]
+}
+
+// xmlDescribeTagsResponse is the wire shape for DescribeTags.
+type xmlDescribeTagsResponse struct {
+	XMLName            xml.Name              `xml:"DescribeTagsResponse"`
+	Xmlns              string                `xml:"xmlns,attr"`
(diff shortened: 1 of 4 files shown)


sivchari (changes requested):
## Review

Two issues to fix before merge:

### 1. `tagsNoOp` returns `<AddTagsResponse>` for both AddTags and RemoveTags

The `tagsNoOp` handler hardcodes `XMLName: "AddTagsResponse"`, so when called for `RemoveTags`, the response XML root element is wrong. Split into separate handlers or parameterize the XML name so `RemoveTags` returns `<RemoveTagsResponse>`.

### 2. XML types should be in `types.go`, not `handlers.go`

Per project convention, request/response types belong in `types.go`. The inline XML structs (`xmlDescribeTagsResponse`, `xmlDescribeCapacityReservationResponse`, `xmlDescribeListenerAttrResponse`, etc.) should be moved there.

sivchari (approved):
Review comments addressed: tagsNoOp split into separate handlers, XML types moved to types.go. LGTM.

PR #569, opened 2026-05-08
feat(cloudcontrol): AWS Cloud Control API + S3 Bucket / EC2 VPC+Subnet / IAM Role

mizchi:
## Summary

Adds AWS Cloud Control API as a JSON 1.0 service in kumo. Cloud Control is the unified CRUD wire interface AWS exposes for any CloudFormation-modelled resource type. Implementing it gives kumo a single, uniform CRUD entrypoint that grows by adding handlers, not by re-deriving each per-service AWS API.

**Verified end-to-end with \`terraform-provider-awscc\`** — \`tofu apply\` / \`tofu plan\` (no drift) / \`tofu destroy\` all succeed against the four registered resource types in a single HCL stack.

## What's in this PR

- New service: \`internal/service/cloudcontrol\`
  - All six Cloud Control operations: \`CreateResource\` / \`GetResource\` / \`UpdateResource\` / \`DeleteResource\` / \`ListResources\` / \`GetResourceRequestStatus\`
  - Synchronous execution: every operation runs against the underlying storage immediately, then returns a synthetic \`ProgressEvent\` with \`OperationStatus=SUCCESS\`
  - JSON 1.0 dispatch via the existing \`JSONProtocolService\` interface using the \`CloudApiService.<Action>\` X-Amz-Target prefix
  - \`EventTime\` encoded as Unix-epoch seconds (number), not RFC3339 string — the AWS JSON 1.0 protocol's timestamp decoder rejects strings.
  - \`progressTracker\` holds per-\`RequestToken\` events; \`GetResourceRequestStatus\` echoes the original \`Identifier\` + \`TypeName\` so the awscc provider's polling loop can follow up with \`GetResource\`. Without this echo the provider 400s on \`GetResource\` and \`(known after apply)\` never resolves.

- A \`Handler\` interface (\`Create\`/\`Read\`/\`Update\`/\`Delete\`/\`List\` on JSON-bytes) + \`Registry\` mapping \`AWS::<Service>::<Resource>\` → handler. Per-resource handler files self-register via \`init()\` → \`registerDefaultHandler\`.

- Generic \`lookupStorage[T](serviceName)\` helper so each handler is one short file: \"give me \`s3.Storage\` / \`ec2.Storage\` / \`iam.Storage\`\" without duplicating registry-walk + interface-assert boilerplate.

## Resource handlers

Four handlers covering the most common resource types in a typical VPC + IAM stack:

| Type | Maps to | Notes |
|---|---|---|
| \`AWS::S3::Bucket\` | \`s3.Storage\` | Cross-protocol: a bucket created via REST is visible via Cloud Control \`ListResources\`, and vice-versa |
| \`AWS::EC2::VPC\` | \`ec2.Storage\` | \`CidrBlock\` required |
| \`AWS::EC2::Subnet\` | \`ec2.Storage\` | \`VpcId\` and \`CidrBlock\` required |
| \`AWS::IAM::Role\` | \`iam.Storage\` | Accepts \`AssumeRolePolicyDocument\` as either a JSON string or an object and round-trips it as an object |

Each handler emits the **full CloudFormation resource type schema** in its \`ResourceModel\` — properties kumo doesn't model surface as JSON null (or empty array for list-typed fields). Without this, \`terraform-provider-awscc\` rejects every Computed property as \"still indicated an unknown value\" and refuses to write state.

\`s3.Service.Storage()\` / \`ec2.Service.Storage()\` / \`iam.Service.Storage()\` accessors are added so the cloudcontrol handlers can resolve the live storage at request time without an init-order coupling.

## Verified

\`\`\`hcl
provider \"awscc\" {
  endpoints = {
    cloudcontrolapi = \"http://localhost:4566\"
    iam             = \"http://localhost:4566\"
    sts             = \"http://localhost:4566\"
  }
}

resource \"awscc_s3_bucket\"   \"demo\" { bucket_name = \"cc-tf-bucket\" }
resource \"awscc_ec2_vpc\"     \"demo\" { cidr_block  = \"10.55.0.0/16\" }
resource \"awscc_ec2_subnet\"  \"demo\" { vpc_id = awscc_ec2_vpc.demo.vpc_id, cidr_block = \"10.55.1.0/24\" }
resource \"awscc_iam_role\"    \"demo\" { role_name = \"cc-tf-role\", assume_role_policy_document = jsonencode(...) }
\`\`\`

| step | result |
|---|---|
| \`tofu apply\` | Apply complete! Resources: 4 added |
| \`tofu plan -detailed-exitcode\` | exit 0 — no drift |
| \`tofu destroy\` | Destroy complete! Resources: 4 destroyed |

Also verified directly with \`aws cloudcontrol\` CLI — Create / Get / List / Delete all succeed, and CC-created resources co-exist with REST/Query-created ones in the same store.

## Out of scope

- Tags as actual storage (handlers emit empty Tags lists today)
- Sub-resource configurations on S3 (encryption / lifecycle / versioning) — they round-trip as null
- \`Update\` semantics beyond \"echo current state\" (PatchDocument is parsed but ignored)
- \`List\` pagination
- Resource types beyond the four above — adding more is one short file per type via \`registerDefaultHandler\`

## Test plan

- [x] \`go build ./...\` passes
- [x] Unit tests cover the handler lifecycle and error paths
- [x] \`aws cloudcontrol\` CLI lifecycle (Create / Get / List / Delete + cross-protocol bucket sharing)
- [x] \`terraform-provider-awscc\` end-to-end (apply / plan-no-drift / destroy on all four types)


sivchari (changes requested):
## Review

### 1. Integration test is required

Per project convention (CLAUDE.md New Service Checklist), every new service must have an integration test at `test/integration/cloudcontrol_test.go` using the golden test pattern. Please add one.

### 2. `isIAMNotFound` relies on fragile string matching

```go
func isIAMNotFound(err error) bool {
    msg := err.Error()
    return strings.Contains(msg, "NoSuchEntity") ||
           strings.Contains(msg, "does not exist") ||
           strings.Contains(msg, "not found")
}
```

`"not found"` is too generic and risks false positives from unrelated errors in the chain. Introduce a typed error in the IAM package (e.g. `*NotFoundError`) and use `errors.As` instead.

sivchari (approved):
Review comments addressed: integration test added, isIAMNotFound uses errors.As with typed error. LGTM.

PR #573, opened 2026-05-09
fix(server): require path boundary when matching admin route prefixes

mizchi:
## Summary

`extractRoutePrefix` and `Router.ServeHTTP` matched service prefixes (`/kumo`, `/lambda`, …) by raw string-prefix, so a request to `/kumo-audit-bad-bucket` was dispatched to the `/kumo` admin router instead of the S3 wildcard route `PUT /{bucket}`, returning 404. Same shadow potentially affects buckets whose names start with `/iam`, `/lambda`, `/snapshots`, `/jobs`, `/findings`, `/-`, etc.

The fix adds `hasPathPrefix` that requires the next character after the prefix to be `/` (or end-of-string) and uses it from both call sites.

A bucket whose name **equals** an admin prefix exactly (e.g. `PUT /kumo`) remains shadowed — that's an unavoidable conflict until kumo grows a Host- or scheme-based admin discriminator. Documented in the test as a known limitation.

## Repro (before fix)

```
$ kumo &
$ curl -X PUT http://localhost:4566/kumo-audit-bad-bucket
404 page not found
```

## After fix

```
$ curl -X PUT http://localhost:4566/kumo-audit-bad-bucket
HTTP/1.1 200 OK
Location: /kumo-audit-bad-bucket
$ aws --endpoint-url=http://localhost:4566 s3api list-buckets
{ "Buckets": [{ "Name": "kumo-audit-bad-bucket", ... }] }
```

## Test plan

- [x] `go test ./internal/server/` — passes including new `TestRouter_PrefixMatchRespectsBoundary` and `TestExtractRoutePrefix_BoundaryGuard`
- [x] Manual: `curl -X PUT http://localhost:4566/<bucket-name-starting-with-admin-prefix>` returns 200 and `s3api list-buckets` shows it


mizchi:
Superseded by #574 — the virtual-hosted-style support PR already includes this commit and exercises the same end-to-end scenario. Bundled because reviewing them together is easier than two separate roundtrips on the same `Router.ServeHTTP`.

PR #574, opened 2026-05-09
fix(server): path-prefix boundary + AWS S3 virtual-hosted-style support

mizchi:
## Summary

Two related router fixes that together let the AWS terraform provider talk to kumo for S3 (and any future bucket-named admin-prefix collisions). Bundled because they touch the same `Router.ServeHTTP` and share an end-to-end repro: `tofu apply` of `aws_s3_bucket` against kumo. (Replaces #573, which is closed in favour of this PR.)

### Fix 1 — path-prefix boundary

`extractRoutePrefix` and `Router.ServeHTTP` matched service prefixes (`/kumo`, `/lambda`, …) by raw string-prefix, so a request to `/kumo-audit-bad-bucket` was dispatched to the `/kumo` admin router instead of the S3 wildcard route `PUT /{bucket}`, returning 404. Same shadow potentially affects buckets whose names start with `/iam`, `/lambda`, `/snapshots`, `/jobs`, `/findings`, `/-`, etc.

`hasPathPrefix` requires the next character after the prefix to be `/` (or end-of-string) and is used at both call sites. A bucket whose name **equals** an admin prefix exactly (e.g. `PUT /kumo`) remains shadowed — that's an unavoidable conflict until kumo grows a Host- or scheme-based admin discriminator. Documented in the test as a known limitation.

### Fix 2 — S3 virtual-hosted-style host headers

The aws-sdk-go-v2 (and therefore the terraform AWS provider) defaults to **virtual-hosted-style** for S3: the bucket lives in the `Host` header (`<bucket>.localhost:4566` for kumo / `<bucket>.s3.amazonaws.com` for real AWS) and the URL path is just `/`. kumo only knew path-style (`/{bucket}`), so:

- `HEAD /` matched the wildcard with `bucket=""` and returned 200 → the SDK reads that as "bucket exists" → terraform reports a spurious `BucketAlreadyExists` when it tries to create the bucket.
- `PUT /` (CreateBucket) was 405.
- Object ops on `<bucket>.localhost/key.txt` couldn't reach `/{bucket}/{key...}` either.

A host-rewrite step at the start of `Router.ServeHTTP` rewrites the request to path-style (`/<bucket>` / `/<bucket>/<key>`) when the Host has the shape:

```
<bucket>.localhost(:port)
<bucket>.s3.amazonaws.com
<bucket>.s3.<region>.amazonaws.com    (incl. dualstack subdomains)
<bucket>.s3-<region>.amazonaws.com    (legacy dash form)
```

Hosts without a bucket prefix (`s3.amazonaws.com`, `localhost:4566`, raw IP) are left untouched.

## End-to-end repro

Before this PR:

```
$ terraform apply (resource = aws_s3_bucket: kumo-audit-bad-bucket)
Error: creating S3 Bucket (kumo-audit-bad-bucket): operation error S3:
       CreateBucket, https response error StatusCode: 404, ...
       NotFound: Not Found
```

After:

```
$ tofu apply -target=aws_s3_bucket.insecure
aws_s3_bucket.insecure: Creation complete after 0s [id=kumo-audit-bad-bucket]
Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
```

## Test plan

- [x] `go test ./internal/server/` — passes including new tests:
  - `TestRouter_PrefixMatchRespectsBoundary`
  - `TestExtractRoutePrefix_BoundaryGuard`
  - `TestRouter_VirtualHostedStyleS3` (9 sub-cases)
  - `TestExtractBucketFromHost` (13 sub-cases)
- [x] Manual: `tofu apply` of an `aws_s3_bucket` against a kumo built from this branch creates the bucket, and `aws s3api list-buckets` returns it

### DIFF
--- a/internal/server/router.go
+++ b/internal/server/router.go
@@ -6,6 +6,7 @@ import (
 	"log/slog"
 	"net/http"
 	"path"
+	"strings"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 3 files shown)


sivchari (approved):
LGTM

PR #572, opened 2026-05-08
fix(ec2): security group ingress/egress rules + DescribeSecurityGroups

mizchi:
## Summary

\`terraform-aws\`'s \`aws_security_group\` lifecycle was effectively broken end-to-end against kumo. Three independent gaps that all surface together:

1. **\`AuthorizeSecurityGroupIngress\` / \`AuthorizeSecurityGroupEgress\` stored *empty* rule sets** because the form-to-JSON converter doesn't expand \`IpPermissions.N.{IpProtocol,FromPort,ToPort,IpRanges.M.CidrIp}\` into a slice — the dotted keys land at the top level of the JSON envelope and never unmarshal into \`req.IPPermissions\`. Same bug class as the prior \`RegisterTargets\` / \`CreateListener\` / \`CreateRule\` fixes.

2. **\`RevokeSecurityGroupIngress\` / \`RevokeSecurityGroupEgress\` weren't implemented at all.** terraform-aws issues these on every Create to clear the default \`0.0.0.0/0\` egress rule before applying user rules; without a 200 response the resource fails to create and downstream Authorize calls never happen.

3. **\`DescribeSecurityGroups\` wasn't implemented.** Even if (1) and (2) worked, the ingress / egress state was unobservable from the API — meaning drift detection, audit tooling, and any second-pass terraform plan all see an empty SG.

How I caught it: the \`kumo-audit\` PoC (post-apply security review) tried to flag SGs open to \`0.0.0.0/0\` and got \`<ipPermissions></ipPermissions>\` back from \`DescribeSecurityGroups\`. Walking back from there surfaced all three gaps.

## What this PR does

- **Parses \`IpPermissions.N.X\`** from \`r.Form\` in both Authorize handlers when the JSON-decoded slice is empty (mirrors the existing fixes on RegisterTargets / CreateListener / CreateRule).
- **Adds \`RevokeSecurityGroupIngress\` / \`RevokeSecurityGroupEgress\`** as no-op success handlers (kumo doesn't track egress modifications, so the no-op is behaviourally equivalent to the cleanup terraform expects).
- **Adds \`DescribeSecurityGroups\` end-to-end**: storage method, handler that filters by \`GroupId.N\` / \`GroupName.N\`, the AWS-published XML wire shape (\`item\` / \`ipPermissions\` / \`item\` / \`ipRanges\` / \`item\`) plus dispatcher and Actions list registration.
[...]

PR #575, opened 2026-05-09
feat(ec2): real RevokeSecurityGroupIngress / RevokeSecurityGroupEgress

mizchi:
## Summary

Replace the no-op `RevokeSecurityGroup{Ingress,Egress}` stub from #572 with a real per-CIDR implementation that mutates storage. terraform sends \`Revoke\` whenever \`cidr_blocks\` shrinks on an existing SG; the stub left stale rules in kumo state, so anything that audits the SG afterwards (kumo-audit, IaC drift detection, …) saw rules that no longer existed in the source-of-truth config.

## AWS revoke semantics

- Rule identity key is **(IPProtocol, FromPort, ToPort)** — CIDR is *not* part of the key
- Within a matching rule, only the listed CidrIPs are removed (so `Revoke {tcp,22,22,[0.0.0.0/0]}` against an authorized `{tcp,22,22,[0.0.0.0/0, 10.0.0.0/8]}` leaves only `10.0.0.0/8`)
- A rule whose IPRanges become empty is dropped entirely
- A missing SG returns `InvalidGroup.NotFound`
- A missing rule/CIDR is a silent no-op (real AWS returns `InvalidPermission.NotFound`; matches the prior stub's behaviour and terraform doesn't distinguish)

## Implementation

- Storage interface gains `RevokeSecurityGroupIngress` / `RevokeSecurityGroupEgress`.
- Pure helpers (\`revokeMatching\`, \`sameRuleKey\`, \`removeCIDRs\`) keep the storage methods short and unit-testable.
- Handlers re-use the same form-parse recovery path \`Authorize*\` uses to work around kumo's form-to-JSON dotted-key limitation.

## Dependency on #572

Branched off \`fix/ec2-sg-ippermissions-form-parse\`. Whichever merges first, the other can rebase to drop the duplicate. The combination is what makes \`tofu apply -target=aws_security_group\` work end-to-end against kumo (and what lets kumo-audit's SG rule pack actually evaluate).

## Test plan

- [x] \`go test ./internal/service/ec2/ -run TestRevokeSecurityGroup\` — 4 sub-tests pass:
  - \`RemovesMatchingCIDR\` (per-CIDR semantics)
  - \`DropsRuleWhenAllCIDRsRemoved\` (empty-IPRanges → rule deletion)
  - \`Egress\` (egress mirror)
  - \`NotFound\` (InvalidGroup.NotFound on missing SG)
- [x] \`make lint\` — clean
- [x] Manual: \`tofu apply -target=aws_security_group.wide_open\` against this branch creates the SG and \`kumo-audit\` reports \`sg-open-ssh\` / \`sg-open-rdp\` / \`sg-open-all-protocols\` findings against it

### DIFF
--- a/internal/service/ec2/handlers.go
+++ b/internal/service/ec2/handlers.go
@@ -267,6 +267,18 @@ func (s *Service) AuthorizeSecurityGroupIngress(w http.ResponseWriter, r *http.R
 		return
 	}
 
+	// The form-to-JSON converter doesn't expand
+	// IpPermissions.N.{IpProtocol,FromPort,ToPort,IpRanges.M.CidrIp}
+	// into a slice — the dotted keys land at the top level and never
+	// unmarshal into req.IPPermissions. Re-parse the form directly so
+	// the ingress rules actually reach storage. (Same bug class as the
+	// RegisterTargets / CreateListener / CreateRule fixes.)
+	if err := r.ParseForm(); err == nil {
+		if perms := parseIPPermissionsFromForm(r.Form); len(perms) > 0 {
+			req.IPPermissions = perms
+		}
+	}
+
 	err := s.storage.AuthorizeSecurityGroupIngress(r.Context(), req.GroupID, req.GroupName, req.IPPermissions)
 	if err != nil {
 		handleError(w, err)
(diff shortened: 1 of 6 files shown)


sivchari (approved):
LGTM

Issue #416, opened 2026-04-07
AWS CDK, SST / Terraform Deployments?


shellscape:
Does kumo support AWS CDK or Terraform deployments to the container? AWS SDK support is nice, but Cloudformation and TF compatibility would be very nice.

WarpRat:
We'd definitely find real Cloudformation object creation very useful as part of our lambda integration tests, especially if it supported the AWS SAM transformer. This project looks super interesting, especially with the recent localstack licensing changes.

mizchi:
Tried this end-to-end with `tofu` (OpenTofu 1.11) + terraform-provider-aws 5.x against `kumo serve` on `localhost:4566` to see what the gap actually is today.

Fixture (5 resources):

```hcl
resource "aws_s3_bucket"      "demo"  { bucket = "tofu-kumo-demo" }
resource "aws_s3_object"      "hello" { bucket = aws_s3_bucket.demo.id; key = "hello.txt"; content = "hello from tofu" }
resource "aws_sqs_queue"      "jobs"  { name = "tofu-kumo-jobs" }
resource "aws_dynamodb_table" "items" { name = "tofu-kumo-items"; billing_mode = "PAY_PER_REQUEST"; hash_key = "id"; attribute { name = "id"; type = "S" } }
resource "aws_iam_role"       "task"  { name = "tofu-kumo-task"; assume_role_policy = jsonencode({...ecs-tasks...}) }
```

Results:

| resource | result |
|---|---|
| `aws_s3_bucket` | ok |
| `aws_s3_object` | ok |
| `aws_sqs_queue` | ok (~25 s — provider polled `GetQueueAttributes` for what looked like attribute propagation, worth tracing later) |
| `aws_dynamodb_table` | table is created, but apply errors during refresh — `ListTagsOfResource` and `DescribeContinuousBackups` are unimplemented and return 400 |
| `aws_iam_role` | fails — `CreateRole` is unreachable from the unified `/` endpoint because the IAM service is not registered with the AWS Query dispatcher (only `/iam/` direct routing works; terraform sends to `/`) |

Two concrete bugs found in one apply, so the integration-test idea seems valuable.

I'd like to send a small series of PRs:

[...]

PR #578, opened 2026-05-09
feat(iam): UpdateRole / UpdateAssumeRolePolicy / TagRole + unified Query dispatcher

mizchi:
## Summary

Two related IAM gaps surfaced by running real provider stacks against kumo:

1. **CRUD parity** — terraform `aws_iam_role`, pulumi-aws `Role`, and alchemy's `Role` all do `CreateRole` on first apply, then on every subsequent apply they `GetRole` and unconditionally call `UpdateRole` + `UpdateAssumeRolePolicy` + `TagRole` to reconcile `description` / `session-duration` / trust-policy / tag drift. Without these three actions, second-and-later applies fail with `InvalidAction`.
2. **Single-endpoint routing** — terraform-provider-aws sends every API call to the unified `/` endpoint regardless of service. IAM was only registered under `/iam/` and did not implement `QueryProtocolService`, so the Query dispatcher returned `UnknownService` for IAM actions hit via `/`. Discovered while doing the experiment in #416 / #589 — `CreateRole` from the AWS provider failed before any of the actions in (1) could even be tried.

### CRUD (UpdateRole / UpdateAssumeRolePolicy / TagRole)

- **UpdateRole** — patches `Description` and `MaxSessionDuration`. Both are `*string` / `*int` on the storage layer; `nil` means \"leave unchanged\", so terraform-style \"only send what changed\" works.
- **UpdateAssumeRolePolicy** — replaces the trust policy verbatim. Same opaque-JSON treatment `PutRolePolicy` already gives inline policies.
- **TagRole** — merge-by-key upsert: existing keys overwrite, new keys append (matches AWS).

All three return `NoSuchEntity` for a missing role and are wired into the action dispatch map. New types (`UpdateRoleResponse`, `UpdateAssumeRolePolicyResponse`, `TagRoleResponse`) match the empty-result-body shape AWS returns.

### Unified dispatcher (terraform compat)

- Implements `TargetPrefix() / ServiceIdentifier() / Actions() / QueryProtocol()` on the IAM service so `internal/server/server.go` picks it up at registration time and routes IAM actions arriving at `/` to the IAM handlers.
- Fixes `getJSONValue` to restore `r.Body` after each read. The Query dispatcher rewrites the body to a single-shot JSON reader (formToJSON) and IAM handlers call `getFormValue` once per parameter; without restoration the second call would return \"\" and `CreateRole` would surface `AssumeRolePolicyDocument is required` even when the field was sent. (This was the next failure mode after the routing fix.)
- New `dispatcher_test.go`:
  - `TestDispatchAction_JSONBody` — `CreateRole` via JSON body (the post-formToJSON shape), confirms RoleName + AssumeRolePolicyDocument + Description all read through.
  - `TestQueryProtocolService` — verifies the marker methods and that `Actions()` covers the core role lifecycle.
  - `TestGetJSONValue_BodyRestoration` — locks in the contract that multiple reads on a JSON-body request return the right values.

## Test plan

- [x] `go test ./internal/service/iam/...` — green (existing UpdateRole/TagRole/UpdateAssumeRolePolicy tests + the new dispatcher tests, 9 funcs / 11 sub-tests).
- [x] `go test ./...` — full suite green.
- [x] `make lint` — IAM package clean.
- [x] End-to-end: tofu (1.11.6) `apply` and `destroy` of `aws_iam_role` against `kumo serve` on `localhost:4566` — both succeed; before this PR `apply` failed with `UnknownAction` from the dispatcher and after only adding the dispatcher methods (without the body-restoration fix) failed with the spurious \"AssumeRolePolicyDocument is required\".

## Validated against

- alchemy `@sst/alchemy` v0.93.7 — `Role` resource now succeeds against kumo (was failing on the immediate post-Create UpdateAssumeRolePolicy / UpdateRole / TagRole calls).
- terraform / opentofu — `aws_iam_role` now succeeds against kumo via the unified `/` endpoint (was failing at `CreateRole` before, regardless of the CRUD additions).

Cross-ref: #416, #589.

### DIFF
--- a/internal/service/iam/dispatcher_test.go
+++ b/internal/service/iam/dispatcher_test.go
@@ -1,0 +1,113 @@
+package iam
+
+import (
+	"encoding/xml"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+// TestDispatchAction_JSONBody covers the path where the AWS Query dispatcher
+// has converted the form-encoded request to JSON before calling DispatchAction.
+//
+// terraform-provider-aws and other clients that target the unified `/`
+// endpoint hit this code path; the IAM-direct `/iam/` route keeps the form
+// payload as-is. The handlers must accept both shapes.
+func TestDispatchAction_JSONBody(t *testing.T) {
+	t.Parallel()
+
+	svc := New(NewMemoryStorage())
+
+	body := strings.NewReader(`{"RoleName":"json-role","AssumeRolePolicyDocument":"{\"Version\":\"2012-10-17\"}","Description":"from json"}`)
+	req := httptest.NewRequest(http.MethodPost, "/", body)
+	req.Header.Set("Content-Type", "application/x-amz-json-1.0")
+	req.Header.Set("X-Amz-Target", "iam.CreateRole")
+
+	w := httptest.NewRecorder()
+	svc.DispatchAction(w, req)
+
+	if w.Code != http.StatusOK {
+		t.Fatalf("DispatchAction CreateRole (JSON): got %d, body=%s", w.Code, w.Body.String())
+	}
+
+	var resp CreateRoleResponse
+	if err := xml.Unmarshal(w.Body.Bytes(), &resp); err != nil {
+		t.Fatalf("xml unmarshal: %v", err)
+	}
+
+	if got, want := resp.CreateRoleResult.Role.RoleName, "json-role"; got != want {
+		t.Fatalf("RoleName: got %q, want %q", got, want)
+	}
+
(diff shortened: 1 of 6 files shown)


mizchi:
Cross-ref: filed [alchemy-run/alchemy#1404](https://github.com/alchemy-run/alchemy/issues/1404) for the alchemy-side gap (their `aws4fetch`-based EC2 helpers don't honour `AWS_ENDPOINT_URL`). Once that and this PR both land, alchemy → kumo works end-to-end for both IAM Role and EC2 — not just S3.

sivchari (approved):
LGTM

PR #579, opened 2026-05-09
Proposal: env-var endpoint passthrough for RDS (and a model for ElastiCache / AMP / MSK / …)

mizchi:
## Proposal

This PR is a **draft** to discuss an approach before committing to a wider change. Concrete code is RDS-only; the proposal covers the broader pattern.

### Problem

For services kumo emulates today (S3, IAM, EC2, …) the audit / IaC-test loop is closed because both control plane and data plane fit inside kumo. For another class of services this isn't realistic:

- **RDS / Aurora** — can't reimplement Postgres / MySQL wire protocol
- **ElastiCache** — can't reimplement RESP at fidelity
- **Amazon Managed Prometheus / Grafana** — too much surface
- **MSK** — Kafka protocol is a project of its own
- **OpenSearch** — same

For these, callers usually *already have* a real local instance — `docker run postgres`, `redis-server`, `prometheus`, `grafana`, … What they miss is the AWS control-plane in front of it: `aws_db_instance.endpoint` returning a hostname that points at the local thing, so the application stack reads its endpoint exactly the way it would in production.

### Approach

Keep kumo strictly control-plane for these services, and let the developer redirect the returned `Endpoint.Address` / `Endpoint.Port` via env vars to whatever they have running locally. kumo never sees data-plane traffic.

```mermaid
sequenceDiagram
  participant tf as terraform / pulumi
  participant kumo
  participant pg as real Postgres on localhost:5433
  tf->>kumo: CreateDBInstance(engine=postgres, …)
  Note right of kumo: Endpoint resolved from<br/>KUMO_RDS_ENDPOINT_ADDRESS_POSTGRES<br/>KUMO_RDS_ENDPOINT_PORT_POSTGRES
  kumo-->>tf: Endpoint = 127.0.0.1:5433
  tf->>pg: connect (reads aws_db_instance.endpoint)
  pg-->>tf: SELECT 42, version()
```

### What this PR contains (RDS only)

```
KUMO_RDS_ENDPOINT_ADDRESS_<ENGINE>   per-engine address override
KUMO_RDS_ENDPOINT_PORT_<ENGINE>      per-engine port override
KUMO_RDS_ENDPOINT_ADDRESS            global fallback
KUMO_RDS_ENDPOINT_PORT               global fallback
```

- Per-engine wins over global. Engine names with dashes (`aurora-postgresql`) normalise to underscores in the env var key.
- Without any override, behaviour is **byte-identical** to today (AWS-shaped hostname, engine default port).
- One small helper `internal/service/rds/endpoint.go`, used from `CreateDBInstance` and `CreateDBCluster`.

### Verified end-to-end

```sh
docker run -d -p 5433:5432 postgres:17                 # real Postgres
KUMO_RDS_ENDPOINT_ADDRESS_POSTGRES=127.0.0.1 \
KUMO_RDS_ENDPOINT_PORT_POSTGRES=5433 \
  kumo                                                   # kumo control plane
tofu apply ...                                           # creates aws_db_instance
psql -h $(curl ... | extract address) -p 5433 -d app -c 'SELECT version()'
# → real Postgres responds.
```

(There's a separate, mechanical issue: the existing `DescribeDBInstances` response is missing `EngineVersion` / `LicenseModel` / `OptionGroupMemberships` / `DBSubnetGroup`, which makes `tofu apply` itself fail with `Provider produced inconsistent result after apply`. The endpoint passthrough above is independent — verified via `aws rds describe-db-instances` instead. Happy to do that completeness pass in a follow-up if this proposal lands.)

### Open questions for the maintainer

1. **Naming**. `KUMO_RDS_ENDPOINT_ADDRESS_POSTGRES` is verbose but mirrors the AWS SDK's `AWS_ENDPOINT_URL_<SERVICE>` precedent. An alternative shape would be a single `KUMO_RDS_ENDPOINTS=postgres=127.0.0.1:5433,mysql=127.0.0.1:3306` config var. Easier to extend; harder to override piecemeal.
2. **Per-instance vs per-engine**. Currently the override is per-engine. A more granular shape — keyed on the DBInstanceIdentifier — would let the same kumo serve multiple Postgres at different ports. Probably overkill for a first cut.
3. **Generalisation**. If this lands, would you accept follow-up PRs that add the same helper to ElastiCache (`KUMO_ELASTICACHE_ENDPOINT_*`), Amazon Managed Prometheus (`KUMO_AMP_ENDPOINT`), Amazon Managed Grafana (`KUMO_AMG_ENDPOINT`), MSK, OpenSearch? The pattern is the same; the diff per service is ~30 lines.
4. **Resource-level prop instead of env var**. A future option could let the IaC author write `endpoint_override = "127.0.0.1:5433"` directly on `aws_db_instance` in their fixture. That would mean kumo accepting a custom request param outside the AWS shape — feels wrong, but mentioning it as a counterpoint.

### Test plan

- [x] `go test ./internal/service/rds/ -run Endpoint` — 5 unit tests pass
- [x] `make lint` — clean
- [x] Manual E2E with docker postgres on :5433 — `psql` via the kumo-reported endpoint executes real SQL

### DIFF
--- a/internal/service/rds/storage.go
+++ b/internal/service/rds/storage.go
@@ -166,6 +166,8 @@ func (m *MemoryStorage) buildDBInstance(input *CreateDBInstanceInput) *DBInstanc
 		availabilityZone = defaultRegion + "a"
 	}
 
+	endpointAddr, endpointPort := endpointFor(input.Engine, input.DBInstanceIdentifier, m.getDefaultPort(input.Engine))
+
 	instance := &DBInstance{
 		DBInstanceIdentifier:       input.DBInstanceIdentifier,
 		DBInstanceClass:            input.DBInstanceClass,
@@ -188,11 +190,7 @@ func (m *MemoryStorage) buildDBInstance(input *CreateDBInstanceInput) *DBInstanc
 		DeletionProtection:         input.DeletionProtection,
 		Tags:                       input.Tags,
 		VpcSecurityGroups:          buildVpcSecurityGroups(input.VpcSecurityGroupIDs),
-		Endpoint: func() *Endpoint {
-			addr, port := endpointFor(input.Engine, input.DBInstanceIdentifier, m.getDefaultPort(input.Engine))
-
-			return &Endpoint{Address: addr, Port: port}
-		}(),
+		Endpoint:                   &Endpoint{Address: endpointAddr, Port: endpointPort},
 	}
 
 	return instance
--- a/internal/service/rds/storage.go
+++ b/internal/service/rds/storage.go
@@ -188,10 +188,11 @@ func (m *MemoryStorage) buildDBInstance(input *CreateDBInstanceInput) *DBInstanc
 		DeletionProtection:         input.DeletionProtection,
 		Tags:                       input.Tags,
 		VpcSecurityGroups:          buildVpcSecurityGroups(input.VpcSecurityGroupIDs),
-		Endpoint: &Endpoint{
-			Address: fmt.Sprintf("%s.%s.%s.rds.amazonaws.com", input.DBInstanceIdentifier, generateID(), defaultRegion),
-			Port:    m.getDefaultPort(input.Engine),
-		},
+		Endpoint: func() *Endpoint {
+			addr, port := endpointFor(input.Engine, input.DBInstanceIdentifier, m.getDefaultPort(input.Engine))
+
+			return &Endpoint{Address: addr, Port: port}
+		}(),
 	}
 
 	return instance
@@ -375,6 +376,9 @@ func (m *MemoryStorage) CreateDBCluster(_ context.Context, input *CreateDBCluste
 		port = m.getDefaultPort(input.Engine)
 	}
 
+	clusterAddr, clusterPort := endpointFor(input.Engine, input.DBClusterIdentifier, port)
+	readerAddr, _ := endpointFor(input.Engine, input.DBClusterIdentifier+"-ro", port)
+
 	cluster := &DBCluster{
 		DBClusterIdentifier: input.DBClusterIdentifier,
 		DBClusterArn:        m.dbClusterArn(input.DBClusterIdentifier),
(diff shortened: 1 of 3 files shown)


sivchari (changes requested):
Thanks for the proposal -- the idea of redirecting data-plane endpoints to a real local service makes a lot of sense for services kumo can't fully emulate.

A few design decisions I'd like to align on before merging:

**Env var naming**: I'd like to adopt `KUMO_{SERVICE}_BACKEND` as the standard convention across all services. For RDS this would look like:

```
KUMO_RDS_BACKEND=127.0.0.1:5433
KUMO_RDS_BACKEND=postgres=127.0.0.1:5433,mysql=127.0.0.1:3306
```

When the value has no `=`, it applies to all engines. When it's comma-separated `engine=host:port` pairs, each engine gets its own target. Format is always `host:port` without scheme -- kumo adds `http://` internally when needed.

This replaces the four separate `ADDRESS`/`PORT` env vars with a single one per service, and the same pattern works naturally for other services (`KUMO_AMP_BACKEND`, `KUMO_ELASTICACHE_BACKEND`, etc).

**On the open questions**:
- Per-instance: per-engine is enough, no need for per-identifier granularity.
- Generalisation: absolutely welcome, using the same `KUMO_{SERVICE}_BACKEND` convention.
- Resource-level prop: let's not go there -- kumo shouldn't accept parameters outside the AWS API shape.

internal/service/rds/storage.go
```diff
@@ -188,10 +188,11 @@ func (m *MemoryStorage) buildDBInstance(input *CreateDBInstanceInput) *DBInstanc
 		DeletionProtection:         input.DeletionProtection,
 		Tags:                       input.Tags,
 		VpcSecurityGroups:          buildVpcSecurityGroups(input.VpcSecurityGroupIDs),
-		Endpoint: &Endpoint{
-			Address: fmt.Sprintf("%s.%s.%s.rds.amazonaws.com", input.DBInstanceIdentifier, generateID(), defaultRegion),
-			Port:    m.getDefaultPort(input.Engine),
-		},
+		Endpoint: func() *Endpoint {
```
sivchari: This IIFE is a bit hard to read. Could you extract `addr, port` before the struct literal? There's already a precedent in this function with `storageType`, `allocatedStorage`, etc.

sivchari (approved):
LGTM. The implementation correctly addresses all review comments:

- Single `KUMO_RDS_BACKEND` env var with both global (`host:port`) and per-engine (`engine=host:port`) forms
- Scheme-containing values are properly rejected in `parseBackendEndpoint`
- Per-engine entries take precedence over global, with dash/underscore normalization
- Both `buildDBInstance` and `CreateDBCluster` use the new `endpointFor` helper

Tests cover the key scenarios well. Nice work.

PR #582, opened 2026-05-09
feat(aps): Amazon Managed Prometheus emulation with data-plane passthrough

mizchi:
## Summary

Adds the \`aps\` (Amazon Managed Service for Prometheus) service. Same \"control-plane in kumo, data-plane proxied to a real local backend\" pattern as #579 (RDS endpoint passthrough): the protocol kumo can't reasonably reimplement (Prometheus remote_write + PromQL) goes to whatever Prometheus the developer is already running locally.

## Endpoints

```
Control plane (in-memory):
  POST   /workspaces                        CreateWorkspace
  GET    /workspaces                        ListWorkspaces
  GET    /workspaces/{id}                   DescribeWorkspace
  DELETE /workspaces/{id}                   DeleteWorkspace

Data plane (proxied to KUMO_APS_BACKEND):
  POST /workspaces/{id}/api/v1/remote_write  → backend /api/v1/write
  GET  /workspaces/{id}/api/v1/query         → backend /api/v1/query
  GET  /workspaces/{id}/api/v1/query_range   → backend /api/v1/query_range
```

Without \`KUMO_APS_BACKEND\` the data plane returns \`502 BackendUnavailable\` with an explicit hint pointing at the env var; control plane still works.

\`/workspaces\` is added to the router prefix list (alongside \`/applications\` etc) so the requests dispatch through the prefix sub-mux and don't collide with the S3 wildcard \`/{bucket}/{key...}\`.

## Verified end-to-end

```mermaid
sequenceDiagram
  participant app
  participant kumo
  participant prom as real Prometheus on :9091
  app->>kumo: CreateWorkspace
  kumo-->>app: workspaceId=ws-...
  app->>kumo: POST /workspaces/ws-.../api/v1/remote_write<br/>(snappy + protobuf)
  kumo->>prom: POST /api/v1/write (passthrough)
  prom-->>kumo: 204
  kumo-->>app: 204
  app->>kumo: GET /workspaces/ws-.../api/v1/query?query=metric
  kumo->>prom: GET /api/v1/query?query=metric
  prom-->>kumo: {\"value\":[t,99]}
  kumo-->>app: {\"value\":[t,99]}
```

Real test using \`docker run prom/prometheus --web.enable-remote-write-receiver\`, kumo with \`KUMO_APS_BACKEND=http://127.0.0.1:9091\`, and a tiny Go pusher built on the official \`prometheus/prometheus/prompb\` types:

```
$ go run ./aps-push KUMO=http://127.0.0.1:4566 WORKSPACE_ID=ws-... METRIC=kumo_e2e_probe VALUE=99
status=204
pushed kumo_e2e_probe=99 via http://127.0.0.1:4566/workspaces/ws-.../api/v1/remote_write

$ curl -s 'http://127.0.0.1:4566/workspaces/ws-.../api/v1/query?query=kumo_e2e_probe'
{\"status\":\"success\",\"data\":{\"resultType\":\"vector\",\"result\":[
  {\"metric\":{\"__name__\":\"kumo_e2e_probe\",\"source\":\"kumo-aps-push\"},\"value\":[1778333734.877,\"99\"]}
]}}
```

## Test plan

- [x] \`go test ./internal/service/aps/\` — 4 tests pass:
  - control-plane round-trip (create / describe / delete)
  - data-plane returns 502 + explicit hint when \`KUMO_APS_BACKEND\` is unset
  - data-plane proxies to upstream verbatim (path + body)
  - data-plane returns 404 for an unknown workspace
- [x] \`make lint\` — clean
- [x] Manual E2E with \`docker run prom/prometheus\` — see diagram + transcript above

## Why this matters

Two complementary use cases land with this PR:

1. \"Did the metrics my app emits actually leave the box?\" — application points its remote_write client at the kumo-returned \`prometheusEndpoint\`, kumo proxies, the developer's local Prometheus / Grafana shows the data. No code change between dev and prod.
2. Audit-style tests — fixture creates a workspace, integration test pushes a known metric, asserts via PromQL that it landed. Same kumo, same Prometheus, same shape of dashboard the prod system would render.


sivchari (changes requested):
Thanks for adding AMP support. The reverse proxy approach makes sense here since kumo needs to translate paths between the AWS AMP API and native Prometheus.

**Package name**: the AWS SDK Go v2 package is `github.com/aws/aws-sdk-go-v2/service/amp` and the CLI is `aws amp`, so this should be `amp` rather than `aps`. The endpoint prefix `aps-workspaces.*` is a different thing -- our convention follows the SDK package name.

**Env var**: following the `KUMO_{SERVICE}_BACKEND` convention we're establishing, this should be `KUMO_AMP_BACKEND=host:port` (no scheme -- kumo adds `http://` internally when proxying).

**Missing pieces per project conventions**: `types.go` for request/response types, integration test at `test/integration/amp_test.go`, and the blank import in `main.go` needs to go between `appmesh` and `appsync` alphabetically.

internal/service/aps/handlers.go
```diff
@@ -0,0 +1,194 @@
...
+	rp := httputil.NewSingleHostReverseProxy(target)
+
+	originalDirector := rp.Director
+	rp.Director = func(req *http.Request) {
+		originalDirector(req)
+		req.URL.Path = strings.TrimRight(target.Path, "/") + backendPath
+		req.URL.RawQuery = r.URL.RawQuery
+		req.Host = target.Host
+	}
+
```
sivchari: Creating a new `ReverseProxy` on every request means the underlying transport connection pool is never reused. Consider storing a `*httputil.ReverseProxy` on the `Service` struct, initialized once in `New()` when `backend != ""`.

sivchari (changes requested):
The previous review comments have been mostly addressed (package rename to amp, env var rename, types.go added, integration test added). Two remaining issues:

1. The integration test in `test/integration/amp_test.go` uses hand-written assertions instead of the golden test pattern (`golden.New(t, ...).Assert(...)`). Please follow the existing integration test patterns.

2. `writeAPSError` function name still references the old package name `APS`. Please rename to `writeAMPError` for consistency.

PR #583, opened 2026-05-09
feat(cloudfront): RFC 9111-conformant edge cache layer (Cache-Control / Vary / Range / revalidation / swr)

mizchi:
## Summary

Adds a real edge cache to kumo's CloudFront service. Until now CloudFront was management-API-only — distributions could be created and read back, but nothing ever served traffic. This PR makes `/kumo/cdn/{distributionId}/{path...}` a cached reverse proxy with rule evaluation tracking RFC 9111 (HTTP Caching), RFC 9110 (HTTP Semantics), RFC 9213 (CDN-Cache-Control), and RFC 5861 (stale-while-revalidate / stale-if-error).

## What's new

```
internal/service/cloudfront/cache/   pure rule package, fully unit-tested
internal/service/cloudfront/edge.go  HTTP edge handler over a per-distribution
                                     in-memory cache
```

## Spec coverage

| Capability | RFC | Where |
|---|---|---|
| `Cache-Control: max-age` / `s-maxage` precedence | 9111 §5.2 | `EffectiveTTL` |
| `Cache-Control: no-store` / `private` (do not cache) | 9111 §5.2 | `EffectiveTTL`, `IsCacheable` |
| `Cache-Control: no-cache` (revalidate every request) | 9111 §5.2 | `MustRevalidate` |
| `Cache-Control: must-revalidate` semantics (stale path only) | 9111 §5.2.2.2 | `MustRevalidate` |
| `Expires` header (incl. zero / past = explicit-zero freshness) | 9111 §5.3 | `EffectiveTTL` |
| Heuristic freshness (`Last-Modified` × 10%) | 9111 §4.2.2 | `heuristicTTL` |
| Heuristically-cacheable status set | 9110 §15 | `IsHeuristicallyCacheableStatus` |
| Explicit-freshness allows any status | 9111 §3 | `IsCacheable` |
| `Vary` value normalisation (case + whitespace + token-list) | 9110 §12 / §15 | `varyValueEqual` |
| `Vary: *` disables caching | 9111 §4.1 | `VaryDisablesCache` |
| `If-None-Match` (incl. `*`, weak comparison, list) | 9110 §13.1.2 | `IfNoneMatchSatisfied` |
| `If-Modified-Since` | 9110 §13.1.3 | `IfModifiedSinceSatisfied` |
| Cache→origin revalidation w/ conditional headers | 9111 §4.3 | `revalidate` + `ConditionalHeaders` |
| 304 → refresh validators + TTL on cached entry | 9111 §4.3.4 | `mergeRevalidatedHeaders` |
| `Range: bytes=` (suffix / open / closed; multi → fall-through) | 9110 §14 | `ParseRange` + `writePartialContent` |
| `Age` header carried forward from upstream | 9111 §5.1 | `InitialAge` |
| Request `Cache-Control: no-cache / no-store / only-if-cached` | 9111 §5.2.1 | `EvaluateClient` |
| Request `Cache-Control: max-age / min-fresh / max-stale[=N]` | 9111 §5.2.1 | `EvaluateClient` |
| `CDN-Cache-Control` overrides `Cache-Control` at the edge | 9213 §3 | `mergedCacheControl` |
| `stale-while-revalidate` / `stale-if-error` from `CDN-Cache-Control` | 5861 / 9213 §3.1 | `ReadCDNStaleDirectives` + `kickBackgroundRevalidate` |
| Hop-by-hop header stripping | 7230 §6.1 | `isHopByHopHeader` |
| Non-safe methods (PUT / POST / DELETE / PATCH) bypass cache | 9111 §4.4 | `passthrough` |

## CloudFront-specific design choices

- **CDN-Cache-Control wins.** `mergedCacheControl` reads CDN-Cache-Control first, falls back to Cache-Control. This matches CloudFront's documented behaviour (CDN-Cache-Control is the edge-targeted policy; Cache-Control is what the browser sees).
- **`stale-while-revalidate` only via CDN-Cache-Control.** CloudFront does *not* honour swr / sie when they appear in plain Cache-Control — that header is browser-targeted. RFC 9213 §3.1 endorses the same split. This means the cache-tests.fyi swr scenarios (which use Cache-Control swr) are **expected failures** and stay that way; the same directives in CDN-Cache-Control are honoured.
- **Status-code cacheability split.** With explicit freshness any status is cacheable (RFC 9111 §3); without it only the heuristic-cacheable set (RFC 9110 §15) is stored — that's the small list CloudFront defaults to.
- **Origins**: `CustomOriginConfig` (HTTP/HTTPS to any host) and `S3OriginConfig` are both honoured. S3 origins are routed at kumo's own S3 service in path-style (`KUMO_S3_BACKEND`, default `http://127.0.0.1:4566`) so the proxy stays in-process. `selectOrigin` honours `DefaultCacheBehavior.TargetOriginID`. OAI / OAC signing is out of scope — kumo S3 doesn't enforce auth.

## Conformance against [cache-tests.fyi](https://cache-tests.fyi/)

The official RFC 9111 conformance suite (Mark Nottingham's reference set, used by Cloudflare / Fastly / Varnish for their published compliance matrices). Distribution config in the harness has `DefaultTTL=0` so the cache only stores when the response carries explicit or heuristic freshness — the spec-compliant default.

```
272 / 365 PASS  (74.5%)
```

Top remaining failure buckets:

| count | reason | category |
|---|---|---|
| 25 | `Response 2 does not come from cache` | parser strictness + heuristic-status edge cases |
| 17 | `Response 2 comes from cache` | Cache-Control parser edges (quoted values, spaces around `=`) |
| 9 | `Response 3 comes from cache` | successive-stale handling |
| 5 | `stale-while-revalidate` family | by design (CloudFront fidelity — see above) |
| 4 | `Response 2 status is 200, not 304` | residual revalidation edge cases |
| 4 | Interim responses (102/103) | not implemented |
| 3 | `Response 2 status is 405, not 200` | method validation |

## E2E demo (no cache-tests harness)

```sh
# Counter origin so we can see when it's actually contacted
python3 -c '...HTTP server with Cache-Control: max-age=60 + counter...' &

# kumo built from this branch
kumo &

# Distribution → counter origin
DIST_ID=$(curl -s -X POST :4566/2020-05-31/distribution -d @dist.xml \
            | grep -oE '<Id>[^<]+</Id>' | sed 's/<[^>]*>//g')

# 1st: Miss. 2nd: Hit (origin counter doesn't advance).
$ curl -i :4566/kumo/cdn/$DIST_ID/foo
HTTP/1.1 200 OK
Cache-Control: public, max-age=60
X-Cache: Miss from kumo
hello-2
$ curl -i :4566/kumo/cdn/$DIST_ID/foo
HTTP/1.1 200 OK
Cache-Control: public, max-age=60
X-Cache: Hit from kumo
Age: 0
hello-2
```

## Out of scope (explicit follow-ups)

- **`If-None-Match`/`If-Modified-Since` rounding cases** — handled for the common path; a few edge timestamps still miss.
- **Successive-stale (Response 3 comes from cache)** — needs handling around two consecutive stale-then-fresh transitions.
- **Range slicing edge cases** — multi-range, suffix-of-zero, etc.
- **Interim responses (102/103)** — not implemented.
- **Cache invalidation actions** — the existing `CreateInvalidation` handler doesn't yet evict from this edge cache; trivial wire-up once the API shape settles.
- **Method validation gaps (405)** — distribution-level allowed-methods enforcement.

## Test plan

- [x] `go test ./internal/service/cloudfront/cache/` — covers the precedence table, `Vary: *`, key stability, conditional headers, Range parser, request directives, heuristic freshness.
- [x] `go test ./internal/service/cloudfront/` — edge tests covering hit/miss, no-store, Vary split, s-maxage override, unknown distribution, non-GET passthrough.
- [x] `go test ./...` — full kumo suite green.
- [x] `make lint` — clean.
- [x] Manual E2E with a counter origin — Miss / Hit / Age behave per spec.
- [x] [cache-tests.fyi](https://cache-tests.fyi) suite — 272 / 365 PASS.

### DIFF
--- a/internal/service/cloudfront/edge.go
+++ b/internal/service/cloudfront/edge.go
@@ -114,14 +114,18 @@ func sameVarySignature(a, b *cacheEntry) bool {
 	return true
 }
 
-// Edge handles GET requests routed through `/kumo/cdn/{distId}/{path...}`.
+// Edge handles requests routed through `/kumo/cdn/{distId}/{path...}`.
 // It implements the CloudFront edge caching contract:
 //
 //  1. Resolve the distribution + chosen origin.
-//  2. Look up the cache entry. If fresh and not flagged for forced
-//     revalidation, serve it with `X-Cache: Hit from kumo` and the
-//     standard `Age` header.
-//  3. Otherwise fetch from the origin, evaluate cacheability + TTL via
+//  2. For non-safe methods (PUT / POST / DELETE / PATCH), pass through
+//     to the origin without consulting or storing the cache. RFC 9111
+//     §4.4 requires the cache to invalidate on these too — handled in
+//     a follow-up.
+//  3. For safe methods (GET / HEAD), look up the cache. If fresh and
+//     not flagged for forced revalidation, serve it with
+//     `X-Cache: Hit from kumo` and the standard `Age` header.
+//  4. Otherwise fetch from the origin, evaluate cacheability + TTL via
 //     the rules in `cache/`, store the response when allowed, and
 //     serve with `X-Cache: Miss from kumo`.
 //
@@ -138,16 +142,22 @@ func (s *Service) Edge(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	cfg, ok := edgeCacheConfig(dist)
+	originURL, ok := edgeOriginURL(dist, r.PathValue("path"), r.URL.RawQuery)
 	if !ok {
-		http.Error(w, "distribution missing DefaultCacheBehavior", http.StatusServiceUnavailable)
+		http.Error(w, "distribution has no usable origin", http.StatusServiceUnavailable)
 
 		return
 	}
 
-	originURL, ok := edgeOriginURL(dist, r.PathValue("path"), r.URL.RawQuery)
+	if !isCacheableMethod(r.Method) {
+		s.passthrough(w, r, originURL)
+
+		return
+	}
+
+	cfg, ok := edgeCacheConfig(dist)
 	if !ok {
-		http.Error(w, "distribution has no usable origin", http.StatusServiceUnavailable)
+		http.Error(w, "distribution missing DefaultCacheBehavior", http.StatusServiceUnavailable)
 
 		return
 	}
@@ -168,6 +178,44 @@ func (s *Service) Edge(w http.ResponseWriter, r *http.Request) {
 
 	storeIfCacheable(s.edgeCache, distID, base, r, upstream, cfg)
 	writeUpstream(w, upstream, "Miss from kumo", 0)
+}
+
+// isCacheableMethod returns true for the HTTP methods CloudFront ever
+// considers caching. Everything else passes through.
+func isCacheableMethod(m string) bool {
+	return m == http.MethodGet || m == http.MethodHead
+}
+
+// isHopByHopHeader reports whether the given header is per-hop and
+// must not be propagated through a proxy (RFC 7230 §6.1).
+func isHopByHopHeader(name string) bool {
+	switch http.CanonicalHeaderKey(name) {
+	case "Connection",
+		"Keep-Alive",
+		"Proxy-Authenticate",
+		"Proxy-Authorization",
+		"Te",
+		"Trailers",
+		"Transfer-Encoding",
+		"Upgrade",
+		"Host":
+		return true
+	}
+
+	return false
+}
+
+// passthrough forwards the request body verbatim, returns the response
+// without touching the cache. Used for PUT / POST / DELETE / PATCH.
+func (s *Service) passthrough(w http.ResponseWriter, r *http.Request, originURL string) {
+	upstream, err := forwardOrigin(originURL, r)
+	if err != nil {
+		http.Error(w, "origin fetch failed: "+err.Error(), http.StatusBadGateway)
+
+		return
+	}
+
+	writeUpstream(w, upstream, "Bypass from kumo", 0)
 }
 
 // edgeCacheConfig pulls the [MinTTL, DefaultTTL, MaxTTL] triple out of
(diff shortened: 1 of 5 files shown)


sivchari (changes requested):
Really solid work -- the split between `cache/rules.go` (pure rule evaluation) and `edge.go` (HTTP handler) is clean, and the RFC 9111 coverage is impressive for a first cut. A few things to address before merging:

**Data race in background revalidation**: `kickBackgroundRevalidate` writes to `entry.Header`, `StoredAt`, and `TTL` from a goroutine, while other requests can read the same entry concurrently in `serveFromCache`. This needs synchronization -- either a mutex on `cacheEntry` or copying the entry before serving.

**`sameVarySignature` false positive**: when `a.Vary` and `b.Vary` have the same length but different header names (e.g., `["accept-encoding"]` vs `["accept-language"]`), the function only iterates `a.Vary` keys and looks them up in `b.VaryValues` -- a key miss returns empty string on both sides, so they compare equal. Need to verify the header names themselves match.

**No cache eviction**: `edgeCache.entries` grows without bound. A simple entry count cap or basic LRU would prevent OOM in long-running sessions.

Smaller items that can be follow-ups:
- `http.DefaultClient` has no timeout -- worth storing a dedicated `*http.Client` on the Service
- `no-store` with `MinTTL > 0`: AWS CloudFront applies MinTTL even when origin sends `no-store` -- current implementation returns 0
- `isHopByHopHeader` has "Trailers" but RFC 7230 Section 4.4 defines the header field as "Trailer" (singular)

internal/service/cloudfront/edge.go
```diff
@@ -0,0 +1,821 @@
...
+	}
+
+	return false
+}
+
+// passthrough forwards the request body verbatim, returns the response
+// without touching the cache. Used for PUT / POST / DELETE / PATCH.
+func (s *Service) passthrough(w http.ResponseWriter, r *http.Request, originURL string) {
+	upstream, err := forwardOrigin(originURL, r)
+	if err != nil {
```
sivchari, line 584: This writes to `entry.Header`, `StoredAt`, and `TTL` without holding any lock, but `serveFromCache` reads these fields concurrently from other goroutines. This is a data race -- `-race` would flag it if the test exercised this path.

internal/service/cloudfront/edge.go
```diff
@@ -0,0 +1,821 @@
...
+		upstream, err := revalidateOrigin(originURL, cloned, cond)
+		if err != nil {
+			return
+		}
+
+		if upstream.StatusCode == http.StatusNotModified {
+			mergeRevalidatedHeaders(entry.Header, upstream.Header)
+			entry.StoredAt = time.Now()
+			entry.TTL = cache.EffectiveTTL(entry.Header, cfg, time.Now())
+
```
sivchari, line 397: This only checks that `a.Vary` names exist in `b.VaryValues`, but doesn't verify `b.Vary` contains the same header names. If `a.Vary = ["accept-encoding"]` and `b.Vary = ["accept-language"]`, both map lookups return empty strings and compare equal -- false positive. Since `VaryHeaders()` returns sorted slices, a direct element-wise comparison of the names would fix it.

internal/service/cloudfront/edge.go
```diff
@@ -0,0 +1,821 @@
...
+//     origin extends the entry's freshness; 200 replaces it.
+//  6. On a true miss: fetch, evaluate cacheability + TTL via `cache/`,
+//     store when allowed, and serve with `X-Cache: Miss from kumo`.
+func (s *Service) Edge(w http.ResponseWriter, r *http.Request) {
+	distID := r.PathValue("distributionId")
+
+	dist, err := s.storage.GetDistribution(r.Context(), distID)
+	if err != nil {
+		http.Error(w, "no such distribution: "+distID, http.StatusNotFound)
+
```
sivchari, line 267: No size limit or eviction policy -- this will grow unbounded in long-running processes. A simple max-entries cap would be enough for now.

mizchi:
Vary の指摘について対応しました。

- `sameVarySignature` で Vary の値だけでなく、Vary header 名そのものも同じ順序で一致することを確認するようにしました。
- これにより `Accept-Encoding` と `Accept-Language` のように header 名が違うが値がどちらも空になるケースを、同一 variant と誤判定しないようにしています。
- regression test として `TestEdgeCache_StoreKeepsDifferentVaryHeaderNames` を追加し、異なる Vary header 名の variant が 2 件保持されることを確認しています。

対応 commit: `e35d1ef fix(cloudfront): address edge cache review feedback`
CI: Integration / Lint / Unit / Pinact はすべて pass 済みです。

sivchari (approved):
Review comments addressed: mutex added for cache entry synchronization, sameVarySignature now compares header names, evictOldestLocked added. LGTM.

PR #584, opened 2026-05-09
feat(s3): GetObject Range support + UploadPartCopy

mizchi:
## Summary

Two related S3 holes filled in one PR — the second piece (\`UploadPartCopy\`) reuses the byte-range parsing introduced by the first.

### 1. GetObject Range — RFC 9110 §14 single-range support

S3 clients commonly issue \`Range\` requests for chunked / resumable downloads. kumo's \`GetObject\` ignored the \`Range\` header and always returned the full body, so any client doing range-based access re-read the same bytes redundantly.

| Header | Meaning |
|---|---|
| \`bytes=START-END\` | closed range, end inclusive |
| \`bytes=-N\` | suffix — last N bytes |
| \`bytes=START-\` | open-ended — START through end-of-content |
| \`bytes=0-99,200-299\` | multi-range — falls through to 200 (single-range only; matches real S3, which doesn't serve multi-range either) |
| \`items=0-99\` | wrong unit — falls through to 200 (RFC 9110 §14.1.3) |
| \`bytes=99-\` past EOF | unsatisfiable — 416 with \`Content-Range: bytes */<size>\` |

\`GetObject\` and \`HeadObject\` now also advertise \`Accept-Ranges: bytes\` on every successful response so range-aware clients know they can resume.

```
$ curl -i -H 'Range: bytes=5-10' :4566/range-bucket/file.txt
HTTP/1.1 206 Partial Content
Accept-Ranges: bytes
Content-Length: 6
Content-Range: bytes 5-10/32
ETag: \"5392fad6bee373074dbd3edfae6506f2\"

56789A
```

Range parser lives in \`internal/service/s3/range.go\` as package-private — not shared with the cloudfront/cache package's identical helper since cross-service coupling for ~100 lines wasn't worth the import-cycle / interface contortion.

### 2. UploadPartCopy — copy bytes from existing object into a part

S3 multipart upload already supported uploading parts from request bodies (\`UploadPart\`), but lacked **\`UploadPartCopy\`** — the variant that sources part bytes from another (or the same) S3 object via the \`X-Amz-Copy-Source\` header. This is how AWS SDKs concatenate / re-package large objects without re-uploading from the client side: log-roll-up jobs, multipart re-uploads of objects > 5 GiB, \"merge N existing files into one\" workflows.

```
PUT /<dst-bucket>/<dst-key>?partNumber=N&uploadId=...
X-Amz-Copy-Source: /<src-bucket>/<src-key>
X-Amz-Copy-Source-Range: bytes=START-END        (optional)

→ 200 OK
  <CopyPartResult>
    <LastModified>...</LastModified>
    <ETag>\"...\"</ETag>
  </CopyPartResult>
```

\`X-Amz-Copy-Source-Range\` follows the same \`bytes=START-END\` form as the \`Range\` parser above, but per AWS spec **only closed ranges** are accepted — suffix and open-ended forms are rejected with \`400 InvalidArgument\` (matches real S3).

Cross-bucket copy works. \`parseCopySource\` (the helper that already backed \`CopyObject\`) is reused for source parsing.

| condition | status | code |
|---|---|---|
| missing / malformed \`X-Amz-Copy-Source\` | 400 | InvalidArgument |
| source object not found | 404 | NoSuchKey |
| source bucket not found | 404 | NoSuchBucket |
| \`copy-source-range\` out of bounds | 400 | InvalidArgument |
| upload not found / wrong key | 404 | NoSuchUpload |

\`handleMultipartError\` was extended to map \`ObjectError\` to 404 and \`MultipartError{Code: \"InvalidArgument\"}\` to 400 — both already-existing error types benefit (no other call site changes needed).

## Out of scope

- **Multi-range \`multipart/byteranges\` 206 response** — real S3 doesn't serve multi-range either, so this is a fidelity choice, not a gap. If any consumer wants RFC 9110-strict multi-range, a follow-up could add it behind an env flag.
- **\`UploadPartCopy\` source ETag pre-condition headers** (\`x-amz-copy-source-if-match\`, etc.) — orthogonal to this PR.

## Test plan

- [x] \`go test ./internal/service/s3/\` — clean
  - \`TestParseByteRange\` — 10 cases (closed / suffix / open / clamped / unsatisfiable / multi / wrong-unit / empty)
  - \`TestGetObject_RangeRoundTrip\` — 7 HTTP scenarios incl. 416
  - \`TestGetObject_AdvertisesAcceptRanges\`
  - \`TestParseCopySourceRange\` — 8 forms (closed / suffix-rejected / open-rejected / inverted / wrong-unit / spaces / absent)
  - \`TestUploadPartCopy_HTTPRoundTrip\` — full / ranged / out-of-bounds / malformed-source / source-not-found
  - \`TestUploadPartCopy_RoundTripsThroughComplete\` — 2 ranged copies → \`CompleteMultipartUpload\` → final body equals expected concat
- [x] HeadObject golden updated for new \`Accept-Ranges: bytes\` header
- [x] \`make lint\` — clean
- [x] Manual \`curl -H 'Range: ...'\` round-trip — 206 / 416 / Content-Range shapes confirmed

This PR subsumes #585 (filed separately before consolidation).

### DIFF
--- a/internal/service/s3/handlers.go
+++ b/internal/service/s3/handlers.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/xml"
 	"errors"
+	"fmt"
 	"io"
 	"net/http"
 	"sort"
@@ -281,6 +282,12 @@ func (s *Service) handleObjectPut(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if r.URL.Query().Get("uploadId") != "" && r.URL.Query().Get("partNumber") != "" {
+		if r.Header.Get("X-Amz-Copy-Source") != "" {
+			s.UploadPartCopy(w, r)
+
+			return
+		}
+
 		s.UploadPart(w, r)
 
 		return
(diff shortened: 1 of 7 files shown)


sivchari (approved):
LGTM

PR #577, opened 2026-05-09
feat(s3): PutBucketLogging / GetBucketLogging

mizchi:
## Summary

Real Put/Get for S3 bucket server-access-log configuration. Previously \`GET /{bucket}?logging\` always returned an empty \`<BucketLoggingStatus/>\` from the sub-resource stub and \`PUT /{bucket}?logging\` had no dispatch at all (silently fell through to \`CreateBucket\`), which:

- broke \`terraform aws_s3_bucket_logging\` end-to-end against kumo (refresh saw no logging configured even after apply)
- left audit consumers of access-log configuration (CIS 3.6 \"S3 access logging enabled\") unable to distinguish a logged bucket from an unlogged one — the stub-only path locked everything to \"unlogged\"

## Implementation

- \`MemoryBucket\` gains a \`Logging *BucketLoggingConfig\` (TargetBucket + TargetPrefix). \`TargetGrants\` isn't modelled — terraform doesn't set them and audit doesn't read them.
- PUT with an empty / missing \`LoggingEnabled\` clears the config, matching AWS \"opt out\" semantics.
- GET returns \`LoggingEnabled\` when configured, or \`<BucketLoggingStatus xmlns=\"...\"/>\` when not — same shape real AWS uses to say \"logging disabled\".
- The \`?logging\` branch in \`serveBucketSubresourceStub\` is removed; the new GET handler owns it.

## Status codes

| op | response |
|---|---|
| PUT \`/{bucket}?logging\` (LoggingEnabled body) | 200 OK |
| PUT \`/{bucket}?logging\` (empty body) | 200 OK, logging cleared |
| GET \`/{bucket}?logging\` (configured) | 200 OK + LoggingEnabled element |
| GET \`/{bucket}?logging\` (not configured) | 200 OK + empty status |
| any op on missing bucket | 404 NoSuchBucket |

## Test plan

- [x] \`go test ./internal/service/s3/ -run TestBucketLogging\` — 3 tests / 3 sub-tests pass:
  - \`PutGet\` storage round-trip
  - \`OptOut\` (empty config clears state, GET returns nil)
  - \`HTTP\` (PUT 200, GET verbatim, GET on unconfigured bucket returns empty status)
- [x] \`go test ./...\` — full suite green
- [x] \`make lint\` — clean
- [x] Manual: \`curl PUT /bucket?logging\` with the AWS XML payload persists, subsequent \`GET /bucket?logging\` returns the same payload verbatim

[...]

PR #576, opened 2026-05-09
feat(s3): PutBucketPolicy / GetBucketPolicy / DeleteBucketPolicy

mizchi:
## Summary

Real Put/Get/Delete for S3 bucket policies. Previously \`GET /{bucket}?policy\` always returned \`NoSuchBucketPolicy\` from the sub-resource stub and PUT had no route at all (404), which:

- broke \`terraform aws_s3_bucket_policy\` end-to-end against kumo
- left audit-style consumers of bucket policy unable to evaluate anything (CIS 2.1.1 \"deny non-TLS\", wildcard-Principal detection, …)

## Implementation

- \`MemoryBucket\` gains a \`Policy string\` field. AWS treats the document as opaque JSON for Put/Get — IAM-layer evaluation isn't kumo's concern — so storage just round-trips the bytes.
- Three handlers wire into the existing \`handleBucketPut\` / \`handleBucketGet\` / \`handleBucketDelete\` dispatchers via the \`?policy\` query.
- \`policy\` is removed from \`bucketSubresourceErrorCode\` mapping; the new GET handler owns the NoSuchBucketPolicy response.
- \`writeBucketErrorOrInternal\` adds NoSuchBucketPolicy → 404.

## Status codes

| op | response |
|---|---|
| PUT \`/{bucket}?policy\` (body = JSON) | 204 No Content |
| GET \`/{bucket}?policy\` | 200 + Content-Type: application/json + body verbatim |
| DELETE \`/{bucket}?policy\` | 204 No Content |
| GET on a bucket without a policy | 404 NoSuchBucketPolicy |
| any op on a missing bucket | 404 NoSuchBucket |

## Test plan

- [x] \`go test ./internal/service/s3/ -run TestBucketPolicy\` — 3 tests / 7 sub-tests pass:
  - \`PutGetDelete\` storage round-trip
  - \`Errors\` (NoSuchBucket / NoSuchBucketPolicy)
  - \`HTTP\` (PUT 204, GET 200 verbatim, DELETE 204, post-delete GET 404)
- [x] \`go test ./...\` — full suite green
- [x] \`make lint\` — clean
- [x] Manual: curl PUT/GET/DELETE on \`/{bucket}?policy\` against a kumo built from this branch returns the codes documented above

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #587, opened 2026-05-09
feat(s3): conditional requests + presigned response-* overrides + ListObjectsV1 + Object ACL

mizchi:
## Summary

Four small S3 surface gaps, bundled per the \"one PR per service\" convention. All four touch the same S3 handler / storage files and share a reviewer.

| # | Capability | Why it matters |
|---|---|---|
| 1 | Conditional GetObject / CopyObject (\`If-Match\`, \`If-None-Match\`, \`If-Modified-Since\`, \`If-Unmodified-Since\`, \`x-amz-copy-source-if-*\`) | Cache revalidation + optimistic concurrency. Real SDKs send these on every revalidation. |
| 2 | Presigned response-header overrides (\`?response-content-type=...\`, \`?response-content-disposition=...\`, etc.) | Browser download links served via S3 presigned URLs need to force \`Content-Disposition: attachment; filename=...\`. |
| 3 | ListObjectsV1 (marker-paginated, no \`KeyCount\`) | SDK v1, awscli's \`aws s3 ls\`, and a handful of non-AWS S3 clients still emit V1 — they choke on the V2-only \`<KeyCount>\` element. |
| 4 | Object ACL CRUD (\`PUT/GET /{bucket}/{key}?acl\`) | terraform \`aws_s3_object_acl\`, compliance scans, and Cloudtrail-style audit pipelines all roundtrip ACL. |

## 1. Conditional GetObject / CopyObject

GetObject:
| condition | result |
|---|---|
| \`If-Match\` miss | 412 PreconditionFailed |
| \`If-Unmodified-Since\` older (no If-Match) | 412 |
| \`If-None-Match\` hit | 304 Not Modified |
| \`If-Modified-Since\` same/newer (no If-None-Match) | 304 |
| \`*\` wildcard | always matches (RFC 9110 §13.1.1) |

CopyObject — same matching against the *source* object's ETag / LastModified, mapped to 412 on any mismatch (no 304 path on copy).

Evaluation order matches RFC 9110 §13.1: \`If-Match\` → \`If-Unmodified-Since\` → \`If-None-Match\` → \`If-Modified-Since\`.

## 2. Presigned response-* overrides

\`?response-content-type=...\`, \`?response-content-disposition=...\`, \`?response-cache-control=...\`, \`?response-content-encoding=...\`, \`?response-content-language=...\`, \`?response-expires=...\`. Overrides win over object metadata via a small \`setIfAbsent\` helper inside \`writeObjectResponse\`.

## 3. ListObjectsV1

\`handleBucketGet\` now dispatches:
  - \`?list-type=2\`  → existing \`ListObjects\` (V2)
  - default          → new \`ListObjectsV1\`

V1 specifics:
  - \`marker\` request param starts the page **after** that key
  - response uses \`<Marker>\` / \`<NextMarker>\` (no \`KeyCount\`)
  - \`IsTruncated\` reflects whether more entries remain
  - \`NextMarker\` is set only when \`delimiter\` is in use, matching real S3

## 4. Object ACL CRUD

| Method | Path |
|---|---|
| PUT | \`/{bucket}/{key}?acl\` |
| GET | \`/{bucket}/{key}?acl\` |

Body is the standard \`AccessControlPolicy\` XML; \`x-amz-acl: <canned>\` header is also accepted as an alternative input. XML body wins when both are sent (matches real S3).

Storage holds per-object ACL on \`MemoryBucket.ObjectACLs\` (key → \`*ObjectACL\`). Default fallback when no ACL was ever set is the canned \`private\` ACL, encoded as a single FULL_CONTROL grant for the owner so the response shape is always consistent.

| Error condition | Status | Code |
|---|---|---|
| PutObjectAcl on missing bucket | 404 | NoSuchBucket |
| PutObjectAcl on missing key | 404 | NoSuchKey |
| Malformed AccessControlPolicy XML | 400 | MalformedACLError |

## Out of scope

- ACL **enforcement** at request time (kumo doesn't reject GetObject based on stored grants — same posture as the existing bucket-level ACL stub).
- \`x-amz-grant-*\` request headers as an alternative ACL input shape (rarely used in practice; XML body or canned ACL covers >99% of callers).

## Test plan

- [x] \`go test ./internal/service/s3/\` — clean
  - \`TestEvalGetObjectPreconditions\` — 12 cases covering RFC 9110 §13.1 evaluation order
  - \`TestEvalCopySourcePreconditions\` — 5 cases on the source-side variant
  - \`TestGetObject_ConditionalRequests\` — 4 HTTP-layer scenarios (200 / 304 / 412)
  - \`TestGetObject_ResponseHeaderOverrides\` — Content-Type / Content-Disposition / Cache-Control overrides survive the response writer
  - \`TestListObjectsV1_BasicRoundTrip\` — confirms \`<KeyCount>\` is absent from V1 response
  - \`TestListObjectsV1_MarkerPagination\` — 5 keys, 2-key pages, marker advances
  - \`TestListObjectsV2_StillUsesContinuationToken\` — \`?list-type=2\` keeps the V2 response shape
  - \`TestObjectACL_RoundTripWithCannedHeader\`
  - \`TestObjectACL_RoundTripWithBody\` — multi-grant XML round-trip, including Group grantee
  - \`TestObjectACL_GetOnNonExistentObjectIs404\`
- [x] \`make lint\` — clean

### DIFF
--- a/internal/service/s3/conditional.go
+++ b/internal/service/s3/conditional.go
@@ -1,0 +1,102 @@
+package s3
+
+import (
+	"net/http"
+	"strings"
+	"time"
+)
+
+// preconditionResult is the outcome of evaluating RFC 9110 conditional
+// request headers against an object's ETag + Last-Modified.
+type preconditionResult int
+
+const (
+	preconditionPass         preconditionResult = iota // request should proceed normally
+	preconditionNotModified                            // GET/HEAD → 304 Not Modified
+	preconditionFailed                                 // any → 412 Precondition Failed
+)
+
+// evalGetObjectPreconditions implements RFC 9110 §13.1 evaluation order
+// for GET/HEAD: If-Match → If-Unmodified-Since → If-None-Match →
+// If-Modified-Since.
+//
+// Precondition headers are commonly used for cache revalidation
+// (`If-None-Match`) and optimistic concurrency (`If-Match`).
+func evalGetObjectPreconditions(h http.Header, etag string, lastModified time.Time) preconditionResult {
+	if v := h.Get("If-Match"); v != "" {
+		if !matchesAnyETag(v, etag) {
+			return preconditionFailed
+		}
+	} else if v := h.Get("If-Unmodified-Since"); v != "" {
+		if t, err := http.ParseTime(v); err == nil && lastModified.After(t) {
+			return preconditionFailed
+		}
+	}
+
+	if v := h.Get("If-None-Match"); v != "" {
+		if matchesAnyETag(v, etag) {
+			return preconditionNotModified
+		}
+	} else if v := h.Get("If-Modified-Since"); v != "" {
+		if t, err := http.ParseTime(v); err == nil && !lastModified.After(t) {
+			return preconditionNotModified
+		}
+	}
+
+	return preconditionPass
+}
+
+// evalCopySourcePreconditions implements the AWS S3 CopyObject
+// `x-amz-copy-source-if-*` family (semantically the same as the
+// HTTP-level conditionals above, just on the *source* object). All
+// failures collapse to 412 — there's no "304" path on copy.
+func evalCopySourcePreconditions(h http.Header, etag string, lastModified time.Time) bool {
+	if v := h.Get("X-Amz-Copy-Source-If-Match"); v != "" {
+		if !matchesAnyETag(v, etag) {
+			return false
+		}
+	}
+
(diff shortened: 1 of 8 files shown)


sivchari (approved):
LGTM

sivchari:
Approved, but this PR now has merge conflicts from recent S3 merges (#576, #584). Could you rebase on main? Will merge once conflicts are resolved.

sivchari:
Sorry, conflicted again after #577 merge. Could you rebase once more?

Issue #478, opened 2026-04-27
Expand targets for integration test


sivchari:
I'm thinking of expanding targets for e2e like the terraform and cdk, and then some of sdks for each language.

PR #589, opened 2026-05-10
test(integration): terraform / opentofu provider compatibility scaffold

mizchi:
## Summary

Add a provider-driven integration-test slot under `test/integration/` that runs `tofu` / `terraform` against `kumo` via [`terraform-exec`](https://github.com/hashicorp/terraform-exec) and asserts the resources actually exist in kumo afterwards via the AWS SDK.

The motivation is captured in #416: today only direct SDK calls hit the kumo surface in CI, so any breakage along the AWS provider's own call sequence (CreateBucket → many follow-up reads / refresh calls) only shows up in user reports. A small fixture-per-resource scaffold catches that class of regression cheaply, and confirmed-green services here also serve as regression tests for the underlying actions.

## Implementation

- New file `test/integration/terraform_test.go` (build tag `integration`):
  - `findTofuOrTerraform(t)` returns the absolute path to `tofu` (preferred) or `terraform`, or `\"\"` if neither is on PATH; tests `t.Skip()` in that case.
  - `runTerraformFixture(t, providerTF, mainTF, verify)` writes the two `.tf` bodies into `t.TempDir()`, runs `tf.Init` / `tf.Apply`, registers `tf.Destroy` in `t.Cleanup`, then invokes the caller's verify callback.
  - `providerTF(services...)` parameterises the endpoints block so each fixture only opens routes for the services it touches.
  - Four fixtures, all `t.Parallel()`:
    - `TestTerraform_S3` — `aws_s3_bucket` + `aws_s3_object`.
    - `TestTerraform_S3_MultiResource` — two buckets + two objects (one referencing a sibling), exercising the provider's resource graph.
    - `TestTerraform_CloudWatchLogs` — `aws_cloudwatch_log_group` + `aws_cloudwatch_log_stream`. Implicitly regresses the no-op tag stubs from #566.
    - `TestTerraform_EC2_VPC` — exercises the recent EC2 VPC attribute work (#566 / #ec2-vpc-attrs).
  - `init()` sets `TF_PLUGIN_CACHE_DIR` to `\$XDG_CACHE_HOME/kumo-tofu-plugins` so provider downloads happen once across the package; subsequent inits in fresh `t.TempDir`s reuse the cache.
- `test/go.mod`: adds `github.com/hashicorp/terraform-exec v0.25.2`.
- `.github/workflows/integration-test.yaml`: adds `opentofu/setup-opentofu@v1.0.5` so CI actually exercises the new tests.

## Why opentofu, not terraform

`terraform-exec` is HashiCorp's library and works equally well with either binary. OpenTofu is permissively licensed and matches the project's posture; the scaffold prefers `tofu` when both are on PATH but falls back to `terraform` if only the latter is available, so contributors that already have `terraform` installed locally are not forced to install another binary.

## Coverage strategy / next PRs

The fixtures here only cover services that work end-to-end on `main` today (apply + destroy succeed cleanly). Four further service classes were probed during this PR's experiment and surfaced real bugs that will land as separate PRs:

- IAM (`aws_iam_role`) — needs IAM to be reachable from the unified `/` endpoint and a body-restoration fix on the JSON path. Folded into #578.
- DynamoDB (`aws_dynamodb_table`) — needs `ListTagsOfResource` / `DescribeContinuousBackups` stubs. Sent as #590.
- Secrets Manager (`aws_secretsmanager_secret`) — provider calls `GetResourcePolicy` after CreateSecret; not implemented in kumo today. Future PR.
- EventBridge (`aws_cloudwatch_event_bus`) — provider calls `ListTagsForResource` after CreateEventBus; same shape as the logs/ecr stubs. Future PR.
- SQS — `aws_sqs_queue` apply succeeds but takes ~25 s because the provider polls `GetQueueAttributes`; needs an attribute-shape investigation.

Each will pick up a fixture in this file when the underlying gap is fixed.

Cross-ref: #416, #478.

## Test plan

- [x] `KUMO_PORT=4566 bin/kumo serve &` then `cd test && go test -tags=integration -run TestTerraform_ -count=1 ./integration/...` — all four tests pass locally with OpenTofu 1.11.6, ~33 s wall-clock with `-parallel 4`.
- [x] `golangci-lint run --build-tags=integration` — `terraform_test.go` clean (other files in the package have pre-existing lint debt unrelated to this change).
- [x] Skip path: removed `tofu` / `terraform` from `PATH` for a quick local check — tests report `--- SKIP` and the suite continues.


sivchari (changes requested):
## Review

### `init()` calls `os.Setenv` — global test pollution

```go
func init() {
    os.Setenv("TF_PLUGIN_CACHE_DIR", ...)
}
```

Setting environment variables in `init()` affects all tests in the binary, not just this package. Use `t.Setenv` inside each test function, or move the setup to `TestMain`.

mizchi:
Resolved the merge conflicts against the latest sivchari/kumo main.

What I did:
- Merged the latest upstream main into mizchi/kumo:feat/terraform-integration-tests.
- Preserved the Terraform integration-test fixes while keeping upstream's newer SNS and DynamoDB changes, including SetSubscriptionAttributes and the DynamoDB UpdateTable stub.
- Confirmed the PR merge state is now CLEAN.

Verification:
- go test ./...
- make lint
- mise x opentofu@1.11.6 -- go test -tags=integration ./integration -run TestTerraform -count=1 -v
- GitHub Actions are all passing: Integration tests, Lint, Unit tests, and Validate pinned actions.

sivchari (approved):
Review comments addressed: init() + os.Setenv replaced with t.Setenv. LGTM.

sivchari:
Approved -- the t.Setenv fix looks good.

Rebase merge is failing due to conflicts with recent changes on main. Could you rebase on latest main? Once resolved, this is ready to merge.

sivchari:
It's been conflicting yet.

PR #590, opened 2026-05-10
feat(dynamodb): no-op stubs for ListTagsOfResource / Tag / Untag / DescribeContinuousBackups

mizchi:
## Summary

terraform-provider-aws (and pulumi / CDK) refresh DynamoDB table state after `CreateTable` by calling `ListTagsOfResource` and `DescribeContinuousBackups`. Without these, the very first `terraform apply` of an `aws_dynamodb_table` resource fails after the table is created — kumo replied with `UnknownOperationException` so the plan was unrecoverable in-place (the table existed but state could not be refreshed, blocking destroy as well).

Same shape as #565 (ecr) / #566 (logs): wire-level no-ops so refresh paths complete, with the door open for real persistence later.

## Implementation

- `ListTagsOfResource` returns `{\"Tags\":[]}` — the field must be present even when empty; some clients fail to parse missing-Tags responses.
- `TagResource` / `UntagResource` accept and discard the payload.
- `DescribeContinuousBackups` returns `ContinuousBackupsStatus=DISABLED` (and the same for `PointInTimeRecoveryStatus`) for any existing table and `TableNotFoundException` for missing ones, matching AWS semantics that terraform refresh paths depend on.
- Wire types live in `internal/service/dynamodb/types.go` (covered by the existing tagliatelle exemption for AWS PascalCase JSON), handlers in a new `tag_backup_stubs.go`.

## Test plan

- [x] `go test ./internal/service/dynamodb/...` — green (4 new tests, including TableNotFoundException path).
- [x] `make lint` — clean.
- [x] End-to-end: `tofu apply` + `tofu destroy` of `aws_dynamodb_table { billing_mode = \"PAY_PER_REQUEST\"; hash_key = \"id\"; ... }` against `kumo serve` — both succeed; before this PR `apply` errored on `ListTagsOfResource` after creating the table, and `destroy` then errored on the same call during refresh.

Cross-ref: #416, #589.

### DIFF
--- a/internal/service/dynamodb/handlers.go
+++ b/internal/service/dynamodb/handlers.go
@@ -732,6 +732,12 @@ func (s *Service) actionHandlers() map[string]func(http.ResponseWriter, *http.Re
 		"TransactGetItems":   s.TransactGetItems,
 		"BatchWriteItem":     s.BatchWriteItem,
 		"BatchGetItem":       s.BatchGetItem,
+		// Tag and continuous-backup stubs — see tag_backup_stubs.go.
+		// Required for terraform / pulumi / CDK refresh paths after CreateTable.
+		"ListTagsOfResource":        s.ListTagsOfResource,
+		"TagResource":               s.TagResource,
+		"UntagResource":             s.UntagResource,
+		"DescribeContinuousBackups": s.DescribeContinuousBackups,
 	}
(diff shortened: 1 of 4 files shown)


sivchari (approved):
LGTM

PR #591, opened 2026-05-10
feat(eventbridge): Lambda targets + content filters + tag stubs

mizchi:
## Summary

Two related EventBridge gaps:

1. **Lambda target delivery + AWS content filter patterns** — pulled forward from a long-running local branch. PutEvents now dispatches asynchronously to Lambda targets the same way SQS targets are dispatched, and the event-pattern matcher accepts the AWS content filters (\`prefix\` / \`suffix\` / \`exists\` / \`anything-but\` / \`numeric\` / \`equals-ignore-case\`) plus mixed literal/filter arrays.
2. **Tag stubs** — discovered while running tofu against EventBridge for #589. terraform-provider-aws calls \`ListTagsForResource\` on every refresh of \`aws_cloudwatch_event_bus\` / \`aws_cloudwatch_event_rule\`; without it, \`tofu apply\` of any EventBridge resource fails with \`InvalidAction\` immediately after create.

### Lambda targets + content filters (commit 1)

- \`PutEvents\` resolves rule targets and, for ARNs in the \`lambda\` namespace, performs an asynchronous Invoke against the local Lambda emulator (mirrors the existing SQS dispatch path).
- \`pattern.go\` extended for the standard content-filter operators. Mixed arrays — \`{ \"detail\": { \"k\": [\"literal\", { \"prefix\": \"x\" }] } }\` — work, matching AWS semantics where any element matches.
- New unit tests in \`pattern_test.go\` and \`storage_test.go\`; integration test in \`test/integration/eventbridge_test.go\` covers the SQS + Lambda dispatch path end-to-end.

### Tag stubs (commit 2)

- \`ListTagsForResource\` returns \`{\"Tags\":[]}\` (the field must be present even when empty; some clients fail to parse missing-Tags responses).
- \`TagResource\` / \`UntagResource\` accept and discard the payload.
- Same shape as the \`ecr\` (#565) / \`logs\` (#566) / \`dynamodb\` (#590) tag stubs — wire-level no-ops so refresh paths complete, real persistence layered later.

## Test plan

- [x] \`go test ./internal/service/eventbridge/...\` — green (existing + the cherry-picked pattern_test / storage_test cases).
- [x] \`go test ./...\` — full suite green.
- [x] \`make lint\` — eventbridge package clean.
- [x] End-to-end: tofu \`apply\` and \`destroy\` of \`aws_cloudwatch_event_bus\` against \`kumo serve\` — both succeed; before this PR \`apply\` failed with \`InvalidAction\` on \`ListTagsForResource\` immediately after creating the bus.

Cross-ref: #416, #589.

### DIFF
--- a/internal/service/eventbridge/handlers.go
+++ b/internal/service/eventbridge/handlers.go
@@ -33,6 +33,11 @@ func (s *Service) getActionHandlers() map[string]handlerFunc {
 		"CreateApiDestination":   s.CreateAPIDestination,
 		"DescribeApiDestination": s.DescribeAPIDestination,
 		"DeleteApiDestination":   s.DeleteAPIDestination,
+		// Tag stubs — see tag_stubs.go.
+		// Required by terraform-provider-aws after CreateEventBus.
+		"ListTagsForResource": s.ListTagsForResource,
+		"TagResource":         s.TagResource,
+		"UntagResource":       s.UntagResource,
 	}
 }
 
--- a/internal/service/eventbridge/pattern.go
+++ b/internal/service/eventbridge/pattern.go
@@ -1,14 +1,20 @@
 package eventbridge
 
 import (
+	"bytes"
 	"encoding/json"
 	"strings"
 )
 
 // matchEventPattern checks if an event matches an EventPattern.
 // EventPattern is a JSON object where each key maps to an array of acceptable values.
+// Pattern array elements can be literals (string/number) or content filter objects:
+//
+//	{"prefix": "x"}, {"suffix": "y"}, {"exists": true},
+//	{"anything-but": [...]}, {"numeric": [">", 0, "<", 10]},
+//	{"equals-ignore-case": "abc"}.
+//
 // An event matches if ALL keys in the pattern match.
-// For each key, the event's value must match at least one value in the pattern array.
 func matchEventPattern(patternJSON string, event *PutEventsRequestEntry) bool {
 	if patternJSON == "" {
 		return true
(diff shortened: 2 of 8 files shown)


sivchari (changes requested):
## Review

### 1. `http.Client` created per delivery — connection leak

`deliverToLambda` creates a new `http.Client{Timeout: 5 * time.Second}` on every invocation inside a goroutine. The underlying transport connections are never reused or closed. Store a shared `*http.Client` as a field on `MemoryStorage` or at package level.

### 2. `context.Background()` prevents graceful shutdown

`deliverToLambda` uses `http.NewRequestWithContext(context.Background(), ...)`. On server shutdown, these goroutines will not be cancelled. Propagate a cancellable context (e.g. from the storage or a shutdown signal) instead.

mizchi:
Addressed both review points in 058b488:

1. **Shared `http.Client`** — `MemoryStorage` now holds a single `*http.Client` (created in `NewMemoryStorage`) and all three deliver paths (`deliverToHTTP` / `deliverToSQS` / `deliverToLambda`) use it. Keep-alives now actually re-use the transport instead of leaking a fresh connection pool per event.
2. **Cancellable context** — added `deliveryCtx` / `deliveryCancel` to `MemoryStorage`, scoped via `context.WithCancel(context.Background())` in `NewMemoryStorage`, cancelled by `Close()`. The three `http.NewRequestWithContext` calls now use `s.deliveryCtx`, so in-flight async deliveries abort on shutdown instead of outliving the server.

`go test ./internal/service/eventbridge/...` and `go vet` clean.

PR #592, opened 2026-05-10
feat(secretsmanager): GetResourcePolicy / Put / Delete stubs

mizchi:
## Summary

terraform-provider-aws calls \`GetResourcePolicy\` on every refresh of \`aws_secretsmanager_secret\`. Without it, \`tofu apply\` of any SecretsManager resource fails with \`InvalidAction\` immediately after the secret is created (the secret exists in storage but state cannot be refreshed, blocking destroy as well).

Same shape as #565 (ecr) / #566 (logs) / #590 (dynamodb) — wire-level no-ops so refresh paths complete, with the door open for real persistence later.

## Implementation

- \`GetResourcePolicy\` returns \`ARN\` + \`Name\` with no \`ResourcePolicy\` field when no policy is attached, matching AWS behaviour. Returns \`ResourceNotFoundException\` for missing secrets.
- \`PutResourcePolicy\` / \`DeleteResourcePolicy\` accept and discard the payload.
- Wire types live in \`types.go\` (covered by the existing tagliatelle exemption).

## Test plan

- [x] \`go test ./internal/service/secretsmanager/...\` — green (3 new tests + sub-tests).
- [x] \`make lint\` — clean.
- [x] End-to-end: tofu \`apply\` + \`destroy\` of \`aws_secretsmanager_secret\` against \`kumo serve\` — both succeed; before this PR \`apply\` errored on \`GetResourcePolicy\` after creating the secret.

Cross-ref: #416, #589.

### DIFF
--- a/internal/service/secretsmanager/handlers.go
+++ b/internal/service/secretsmanager/handlers.go
@@ -484,6 +484,12 @@ func (s *Service) DispatchAction(w http.ResponseWriter, r *http.Request) {
 		s.UpdateSecret(w, r)
 	case "GetRandomPassword":
 		s.GetRandomPassword(w, r)
+	case "GetResourcePolicy":
+		s.GetResourcePolicy(w, r)
+	case "PutResourcePolicy":
+		s.PutResourcePolicy(w, r)
+	case "DeleteResourcePolicy":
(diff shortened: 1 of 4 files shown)


sivchari (approved):
LGTM

PR #593, opened 2026-05-10
feat(kms): GetKeyPolicy / ListResourceTags / GetKeyRotationStatus stubs

mizchi:
## Summary

terraform-provider-aws calls \`GetKeyPolicy\`, \`ListResourceTags\`, and \`GetKeyRotationStatus\` on every refresh of \`aws_kms_key\` (and immediately after CreateKey). Without them, \`tofu apply\` of any KMS resource fails with \`InvalidAction\` before the create call returns.

Same shape as #565 / #566 / #590 / #592 — wire-level no-ops so refresh paths complete, with the door open for real persistence later.

## Implementation

- \`GetKeyPolicy\` returns the AWS-default key policy (standard AccountRootEnable statement) so terraform's drift hash stays stable across applies.
- \`ListKeyPolicies\` reports the single \`default\` name AWS exposes.
- \`ListResourceTags\` returns an empty Tags array (the field must be present even when empty so terraform can parse it).
- \`GetKeyRotationStatus\` reports rotation as disabled.
- \`PutKeyPolicy\` / \`TagResource\` / \`UntagResource\` accept and discard their payload.

## Test plan

- [x] \`go test ./internal/service/kms/...\` — green.
- [x] \`make lint\` — clean.
- [x] End-to-end: tofu \`apply\` + \`destroy\` of \`aws_kms_key { description = ...; deletion_window_in_days = 7 }\` against \`kumo serve\` — both succeed; before this PR \`apply\` errored on \`GetKeyPolicy\` after creating the key.

Cross-ref: #416, #589.

### DIFF
--- a/internal/service/kms/handlers.go
+++ b/internal/service/kms/handlers.go
@@ -28,6 +28,15 @@ func (s *Service) getActionHandlers() map[string]handlerFunc {
 		"CreateAlias":         s.CreateAlias,
 		"DeleteAlias":         s.DeleteAlias,
 		"ListAliases":         s.ListAliases,
+		// Key policy + tag stubs — see policy_tag_stubs.go.
+		// Required by terraform-provider-aws after CreateKey.
+		"GetKeyPolicy":         s.GetKeyPolicy,
+		"PutKeyPolicy":         s.PutKeyPolicy,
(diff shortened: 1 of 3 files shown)


sivchari (approved):
LGTM

PR #594, opened 2026-05-10
feat(sfn): ValidateStateMachineDefinition + Versions/Aliases/Tags stubs

mizchi:
## Summary

terraform-provider-aws calls \`ValidateStateMachineDefinition\` during plan and \`ListStateMachineVersions\` / \`ListStateMachineAliases\` / \`ListTagsForResource\` during refresh of \`aws_sfn_state_machine\`. Without them, \`tofu plan\` of any \`aws_sfn_state_machine\` fails with \`InvalidAction\` before reaching the create call, and \`tofu apply\` then errors on the post-create refresh.

Same shape as #565 / #566 / #590 / #592 / #593 — wire-level no-ops so plan/refresh paths complete, with the door open for real persistence later.

## Implementation

- \`ValidateStateMachineDefinition\` reports OK for any input. Definitions are not statically validated in kumo; the real shape check happens in \`CreateStateMachine\`.
- \`ListStateMachineVersions\` / \`ListStateMachineAliases\` return empty arrays (the field must be present even when empty so terraform can parse it).
- \`ListTagsForResource\` returns an empty Tags array. \`TagResource\` / \`UntagResource\` accept and discard their payload.

## Test plan

- [x] \`go test ./internal/service/sfn/...\` — green.
- [x] \`make lint\` — clean.
- [x] End-to-end: tofu \`apply\` + \`destroy\` of a minimal Pass-state \`aws_sfn_state_machine\` against \`kumo serve\` — both succeed; before this PR \`plan\` errored on \`ValidateStateMachineDefinition\`, and after that fix \`apply\` errored on \`ListStateMachineVersions\`.

Cross-ref: #416, #589.

### DIFF
--- a/internal/service/sfn/definition_stub.go
+++ b/internal/service/sfn/definition_stub.go
@@ -1,0 +1,52 @@
+package sfn
+
+import "net/http"
+
+// ValidateStateMachineDefinition reports the supplied definition as valid.
+//
+// terraform-provider-aws calls ValidateStateMachineDefinition during the
+// plan phase of every aws_sfn_state_machine, before issuing CreateStateMachine.
+// Without it, `tofu plan` fails with InvalidAction and the resource never
+// reaches the create path.
+//
+// Definitions are not statically validated in kumo; this stub returns OK
+// so the apply pipeline proceeds and CreateStateMachine does the real
+// shape check.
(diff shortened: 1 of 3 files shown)


sivchari (approved):
LGTM

PR #595, opened 2026-05-10
feat(sns): GetTopicAttributes / SetTopicAttributes / Tag stubs

mizchi:
## Summary

terraform-provider-aws issues \`SetTopicAttributes\` once per attribute it wants to set after \`CreateTopic\`, regardless of whether the resource block specifies them — every \`*FeedbackSampleRate\` (Firehose / Application / Lambda / HTTP / SQS) goes through. Without an accepting handler, the first such call fails with \`InvalidAction\` (the action was missing from the SNS dispatcher) and apply errors before any subsequent attribute is even tried. \`GetTopicAttributes\` and \`ListTagsForResource\` were also missing and would break the post-create refresh.

Found while running tofu against SNS for #589.

## Implementation

- **SetTopicAttributes**: accepts \`AttributeName\` + \`AttributeValue\` and persists them onto the topic's attribute map. The wire field uses a small \`flexString\` helper because the Query→JSON dispatcher promotes \`\"0\"\` / \`\"1\"\` / \`\"true\"\` / \`\"false\"\` to typed JSON literals (\`parseFormValue\` in \`internal/server/awsquery.go\`) but AWS itself sends every \`SetTopicAttributes\` value as a string. Numeric-looking sample-rate values must round-trip as strings, so \`flexString\` accepts string / number / bool and stores the canonical string form.
- **GetTopicAttributes**: returns the topic's stored attributes plus the AWS-managed defaults terraform-provider-aws reads after refresh — \`TopicArn\`, \`Owner\`, \`Policy\` (the standard \`__default_policy_ID\` statement so terraform's drift hash stays stable), \`Subscriptions*\`, \`DeliveryPolicy\` / \`EffectiveDeliveryPolicy\`.
- **ListTagsForResource** / **TagResource** / **UntagResource**: empty / no-op, same shape as the ecr/logs/dynamodb/eventbridge stubs.
- Storage extended with \`GetTopic\` and \`SetTopicAttribute\`. The five new actions are registered with the unified Query dispatcher so they're reachable from \`/\`.

## Test plan

- [x] \`go test ./internal/service/sns/...\` — green.
- [x] \`make lint\` — clean.
- [x] End-to-end: tofu \`apply\` + \`destroy\` of \`aws_sns_topic { name = ...; display_name = \"kumo demo\" }\` against \`kumo serve\` — both succeed; before this PR \`apply\` errored on \`SetTopicAttributes(FirehoseSuccessFeedbackSampleRate)\` immediately after creating the topic.

Cross-ref: #416, #589.

### DIFF
--- a/internal/service/sns/handlers.go
+++ b/internal/service/sns/handlers.go
@@ -460,6 +460,16 @@ func (s *Service) DispatchAction(w http.ResponseWriter, r *http.Request) {
 		s.ListSubscriptions(w, r)
 	case "ListSubscriptionsByTopic":
 		s.ListSubscriptionsByTopic(w, r)
+	case "GetTopicAttributes":
+		s.GetTopicAttributes(w, r)
+	case "SetTopicAttributes":
+		s.SetTopicAttributes(w, r)
+	case "ListTagsForResource":
+		s.ListTagsForResource(w, r)
+	case "TagResource":
+		s.TagResource(w, r)
+	case "UntagResource":
+		s.UntagResource(w, r)
 	default:
 		writeTopicError(w, errInvalidAction, "The action "+action+" is not valid", http.StatusBadRequest)
 	}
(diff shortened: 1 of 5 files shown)


sivchari (approved):
LGTM

PR #596, opened 2026-05-10
fix(ec2): make terraform destroy work for IGW + subnet

mizchi:
## Summary

terraform-provider-aws destroy of \`aws_internet_gateway\` issues \`DetachInternetGateway\` then \`DeleteInternetGateway\`, and destroy of \`aws_subnet\` first calls \`DescribeNetworkInterfaces\` to surface dangling ENIs that would block the delete. None of the three actions were implemented in kumo, so destroy of any IGW or subnet failed with \`InvalidAction\` even when apply had succeeded.

Found while running tofu against EC2 for #589.

## Implementation

- **DetachInternetGateway / DeleteInternetGateway** use the same request shape as Attach / Create. Storage gains:
  - \`DetachInternetGateway\` — pops the matching attachment from the IGW's \`Attachments\` slice; returns \`Gateway.NotAttached\` when not attached, matching AWS.
  - \`DeleteInternetGateway\` — returns \`DependencyViolation\` when still attached, \`InvalidInternetGatewayID.NotFound\` when missing.
- **DescribeNetworkInterfaces** returns an empty \`NetworkInterfaceSet\`. ENIs are not modeled in storage yet; reporting \"no ENIs\" lets destroy proceed for resources whose deletion otherwise blocks on dangling ENIs (load-balancer ENIs, lambda VPC ENIs, etc.). Real ENI emulation is a separate, larger piece of work; this stub unblocks destroy paths for the resources that already work on the create side.

All three actions are wired into both \`getActionHandler\` (direct \`/ec2/\` access) and \`Actions()\` (the unified Query dispatcher served at \`/\`).

## Test plan

- [x] \`go test ./internal/service/ec2/...\` — green.
- [x] \`make lint\` — clean.
- [x] End-to-end: tofu \`apply\` + \`destroy\` of \`aws_vpc\` + \`aws_internet_gateway\` + \`aws_subnet\` against \`kumo serve\` — both succeed cleanly; before this PR \`destroy\` errored on \`DetachInternetGateway\` and \`DescribeNetworkInterfaces\` and the IGW + subnet leaked.

Cross-ref: #416, #589.

### DIFF
--- a/internal/service/ec2/handlers.go
+++ b/internal/service/ec2/handlers.go
@@ -1423,7 +1423,11 @@ func (s *Service) getActionHandler(action string) func(http.ResponseWriter, *htt
 		// Internet gateway operations
 		"CreateInternetGateway":    s.CreateInternetGateway,
 		"AttachInternetGateway":    s.AttachInternetGateway,
+		"DetachInternetGateway":    s.DetachInternetGateway,
+		"DeleteInternetGateway":    s.DeleteInternetGateway,
 		"DescribeInternetGateways": s.DescribeInternetGateways,
+		// Network interface — destroy-time stub
+		"DescribeNetworkInterfaces": s.DescribeNetworkInterfaces,
 		// Route table operations
 		"CreateRouteTable":    s.CreateRouteTable,
 		"CreateRoute":         s.CreateRoute,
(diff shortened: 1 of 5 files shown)


sivchari (approved):
LGTM

PR #597, opened 2026-05-10
feat(route53): ListTagsForResource / ChangeTagsForResource stubs

mizchi:
## Summary

**Required for terraform compatibility.** terraform-provider-aws calls \`ListTagsForResource\` on every refresh of \`aws_route53_zone\` (URL \`GET /2013-04-01/tags/hostedzone/{Id}\`). Without it, \`tofu apply\` of any Route53 zone fails with a 404 right after \`CreateHostedZone\` returns; the zone exists in storage but state cannot be refreshed, so destroy is also blocked.

Found while running tofu against Route53 for #589.

## Implementation

- \`ListTagsForResource\` returns an empty \`Tags\` element wrapped in the standard \`ResourceTagSet\` shape (\`ResourceType=\"hostedzone\"\`).
- \`ChangeTagsForResource\` accepts and discards the payload.
- Routes registered at \`/2013-04-01/tags/{type}/{id}\` (GET + POST), matching the AWS Route53 REST URL layout the SDK uses.
- Tags are not modeled in the storage layer yet. Same shape as the ecr (#565) / logs (#566) / dynamodb (#590) / eventbridge (#591) stubs — wire-level no-ops so refresh paths complete, with the door open for real persistence later.

## Test plan

- [x] \`go test ./internal/service/route53/...\` — green.
- [x] \`make lint\` — clean.
- [x] End-to-end: tofu \`apply\` + \`destroy\` of \`aws_route53_zone { name = \"example.local\" }\` against \`kumo serve\` — both succeed; before this PR \`apply\` errored with a 404 on \`ListTagsForResource\` immediately after creating the zone.

Cross-ref: #416, #589.

### DIFF
--- a/internal/service/route53/service.go
+++ b/internal/service/route53/service.go
@@ -51,6 +51,12 @@ func (s *Service) RegisterRoutes(r service.Router) {
 
 	// Changes
 	r.Handle("GET", "/2013-04-01/change/{id}", s.GetChange)
+
+	// Tags — see tag_stubs.go.
+	// terraform-provider-aws calls these on every refresh of aws_route53_zone;
+	// without them the apply errors on a 404 immediately after CreateHostedZone.
+	r.Handle("GET", "/2013-04-01/tags/{type}/{id}", s.ListTagsForResource)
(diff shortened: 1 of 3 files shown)


sivchari (approved):
LGTM

PR #598, opened 2026-05-10
feat(elbv2): DescribeTags / AddTags / RemoveTags stubs

mizchi:
## Summary

**Required for terraform compatibility.** terraform-provider-aws calls \`DescribeTags\` on every refresh of \`aws_lb\` / \`aws_lb_target_group\` / \`aws_lb_listener\`. Without it, \`tofu apply\` of any ELBv2 resource fails with \`InvalidAction\` immediately after the resource is created and destroy is also blocked.

Found while running tofu against ELBv2 for #589.

## Implementation

- \`DescribeTags\` echoes one \`TagDescription\` per requested \`ResourceArns.member.N\` with an empty Tags list. AWS always returns one entry per requested ARN even when no tags are attached, and the provider crashes if a requested ARN is missing from the response — \`collectResourceArns\` reads both the form-encoded shape (\`ResourceArns.member.1\`, \`.2\`, ...) and the JSON shape produced by the unified Query→JSON dispatcher.
- \`AddTags\` / \`RemoveTags\` accept and discard the payload.
- Tags are not modeled in storage yet. Same shape as the ecr / logs / dynamodb / route53 stubs — wire-level no-op with the door open for real persistence later.

**Drive-by**: removed three stray import-path strings (\`\"github.com/sivchari/kumo/internal/service\"\`, \`\"os\"\`, \`\"io\"\`) from \`Actions()\` that had leaked from a prior diff and were registered as bogus action names with the unified Query dispatcher.

## Test plan

- [x] \`go test ./internal/service/elbv2/...\` — green.
- [x] \`make lint\` — clean.
- [x] End-to-end: tofu \`apply\` + \`destroy\` of \`aws_lb_target_group\` against \`kumo serve\` — both succeed; before this PR \`apply\` errored on \`DescribeTags\` immediately after \`CreateTargetGroup\`.

Cross-ref: #416, #589.

### DIFF
--- a/internal/service/elbv2/handlers.go
+++ b/internal/service/elbv2/handlers.go
@@ -870,6 +870,11 @@ func (s *Service) getActionHandler(action string) func(http.ResponseWriter, *htt
 		"DescribeListeners":              s.DescribeListeners,
 		"ModifyListener":                 s.ModifyListener,
 		"DescribeTargetHealth":           s.DescribeTargetHealth,
+		// Tag stubs — see tag_stubs.go.
+		// Required by terraform-provider-aws after Create*.
+		"DescribeTags": s.DescribeTags,
+		"AddTags":      s.AddTags,
+		"RemoveTags":   s.RemoveTags,
 	}
 
 	return handlers[action]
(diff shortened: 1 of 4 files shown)


sivchari:
Closing: #567 covers the same DescribeTags/AddTags/RemoveTags functionality plus additional stubs (DescribeCapacityReservation, DescribeListenerAttributes, ModifyListenerAttributes). Review comments have been posted on #567 for the fixes needed before merge.

PR #599, opened 2026-05-10
feat(glue): GetTags / TagResource / UntagResource stubs

mizchi:
## Summary

**Required for terraform compatibility.** terraform-provider-aws calls \`GetTags\` on every refresh of \`aws_glue_catalog_database\` / \`aws_glue_catalog_table\` / \`aws_glue_job\`. Without it, \`tofu apply\` of any Glue resource fails with \`Unknown operation\` immediately after the resource is created and destroy is also blocked.

Found while running tofu against Glue for #589.

## Implementation

- \`GetTags\` returns an empty Tags map. Glue uses a Tags map (\`{\"Tags\": {}}\`), not the array-of-{Key,Value} shape AWS uses elsewhere; the field must be present even when empty for the provider to parse it.
- \`TagResource\` / \`UntagResource\` accept and discard the payload.
- Tags are not modeled in the storage layer yet. Same shape as the ecr / logs / dynamodb / route53 / elbv2 stubs — wire-level no-op with the door open for real persistence later.

**Drive-by**: extracted the \`DispatchAction\` switch into an \`actionHandlers()\` map after cyclop tripped on the new cases (15 → 18). Same pattern as the EventBridge / DynamoDB / IAM dispatchers.

## Test plan

- [x] \`go test ./internal/service/glue/...\` — green.
- [x] \`make lint\` — clean.
- [x] End-to-end: tofu \`apply\` + \`destroy\` of \`aws_glue_catalog_database { name = ... }\` against \`kumo serve\` — both succeed; before this PR \`apply\` errored with \`Unknown operation: GetTags\` immediately after \`CreateDatabase\`.

Cross-ref: #416, #589.

### DIFF
--- a/internal/service/glue/handlers.go
+++ b/internal/service/glue/handlers.go
@@ -30,39 +30,35 @@ func (s *Service) DispatchAction(w http.ResponseWriter, r *http.Request) {
 
 	operation := parts[1]
 
-	switch operation {
-	case "CreateDatabase":
-		s.CreateDatabase(w, r)
-	case "GetDatabase":
-		s.GetDatabase(w, r)
-	case "GetDatabases":
-		s.GetDatabases(w, r)
-	case "DeleteDatabase":
-		s.DeleteDatabase(w, r)
-	case "CreateTable":
-		s.CreateTable(w, r)
-	case "GetTable":
-		s.GetTable(w, r)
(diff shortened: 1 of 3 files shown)


sivchari (approved):
LGTM

PR #600, opened 2026-05-10
feat(cloudwatch): register Query dispatcher + XML responses + tag stubs

mizchi:
## Summary

**Required for terraform compatibility.** terraform-provider-aws sends \`PutMetricAlarm\` / \`DescribeAlarms\` / \`DeleteAlarms\` / \`ListTagsForResource\` through the AWS Query protocol over \`/\`. CloudWatch was registered only as a CBOR (RPC v2) service so the unified Query dispatcher returned \`UnknownService\` for every alarm action; even after registration the existing JSON handlers would have returned \`application/x-amz-json-1.0\` bodies which terraform's Query-protocol client cannot parse.

Found while running tofu against CloudWatch for #589.

## Implementation

- **QueryProtocolService on cloudwatch service**: adds \`TargetPrefix\` / \`ServiceIdentifier\` / \`Actions\` / \`QueryProtocol\`. The SDK identifier is \`\"cloudwatch\"\` (not \`\"monitoring\"\`) because terraform-provider-aws sends \`api/cloudwatch#x.y.z\` in the User-Agent — that's the SDK package name, not the AWS service ID.
- **XML response wrappers** (\`xml_query.go\`) for \`PutMetricAlarmResponse\` / \`DeleteAlarmsResponse\` / \`DescribeAlarmsResponse\` with the standard \`ResponseMetadata\` block. The JSON \`DispatchAction\` handlers now write these instead of \`application/x-amz-json-1.0\`. CBOR handlers (\`PutMetricAlarmCBOR\` etc.) are untouched and the existing CBOR integration tests continue to use them.
- **No-op stubs** for \`ListTagsForResource\` / \`TagResource\` / \`UntagResource\` (\`tag_stubs.go\`) — terraform calls \`ListTagsForResource\` on every alarm refresh and apply errors immediately after \`PutMetricAlarm\` without it.

**Drive-by**: \`server/awsquery.go\`'s \`parseServiceFromUserAgent\` now also strips at \`/\` in addition to \`#\` so \`api/<id>/<version>\` User-Agent formats route correctly. Adds a \`ua\` attribute to request logs to make this kind of routing mismatch easier to diagnose (this is what surfaced the \"cloudwatch\" vs \"monitoring\" discrepancy here).

## Test plan

- [x] \`go test ./internal/service/cloudwatch/... ./internal/server/...\` — green (existing CBOR-path tests untouched).
- [x] \`make lint\` — clean.
- [x] End-to-end: tofu \`apply\` + \`destroy\` of \`aws_cloudwatch_metric_alarm\` against \`kumo serve\` — both succeed; before this PR \`apply\` errored on \`PutMetricAlarm\` with \`UnknownService\`, then on \`ListTagsForResource\` after the dispatcher fix.

Cross-ref: #416, #589.

### DIFF
--- a/internal/server/awsquery.go
+++ b/internal/server/awsquery.go
@@ -148,6 +148,10 @@ func (d *QueryProtocolDispatcher) resolveByFallback(action string) (queryHandler
 // parseServiceFromUserAgent extracts the service identifier from the AWS SDK v2 User-Agent header.
 // The User-Agent contains a token like "api/rds#1.5.0"; this function returns "rds".
 // Returns empty string if no api/ token is found.
+//
+// terraform-provider-aws (and a few other SDK consumers) sometimes use "/"
+// instead of "#" as the version separator (e.g. "api/monitoring/1.0"), so
+// both are accepted.
 func parseServiceFromUserAgent(userAgent string) string {
 	for _, token := range strings.Split(userAgent, " ") {
 		after, found := strings.CutPrefix(token, "api/")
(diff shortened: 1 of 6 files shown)


sivchari (approved):
LGTM

PR #601, opened 2026-05-10
fix(lambda): terraform path-extract + refresh stubs + InvokeEndpoint-less stub-mode

mizchi:
## Summary

**Required for terraform compatibility.** terraform-provider-aws sends Lambda calls to the bare \`/2015-03-31/functions/{name}\` path (no \`/lambda/\` prefix). The route registration handled both prefixes, but the path-extraction helper assumed the \`/lambda/\` prefix was always present — for the bare-prefix variant it returned empty and every read errored with \`InvalidParameterValueException: FunctionName is required\`.

After fixing that, the post-create refresh hits more endpoints that were not registered. After all of those, ad-hoc \`aws lambda invoke\` against a terraform-created function still failed because kumo refused to invoke without an explicit \`InvokeEndpoint\` (a kumo-specific extension that no provider sets).

Each piece below was found by running tofu against Lambda for #589 and then exercising the deployed function with the AWS CLI.

### Path-extract for the bare prefix

Rewrites \`extractFunctionName\` and the per-subresource helpers (\`...FromCodePath\`, \`...FromConfigPath\`, \`...FromInvokePath\`) to scan for the \`functions\` segment anywhere in the path. Accepts both \`/lambda/2015-03-31/functions/{name}\` and the bare \`/2015-03-31/functions/{name}\` shape.

### Refresh-path stubs

Adds five stubs that terraform-provider-aws calls right after \`CreateFunction\` (and on every refresh):

- \`ListVersionsByFunction\` returns a single \`$LATEST\` entry built from the function's stored config.
- \`ListAliases\` / \`ListFunctionEventInvokeConfigs\` return empty arrays.
- \`GetFunctionCodeSigningConfig\` returns \`FunctionName\` + no config.
- \`GetPolicy\` returns \`ResourceNotFoundException\` (matches AWS's \"no attached policy\" semantics).

### Router prefix isolation

Adds \`/2019-09-25\` and \`/2020-06-30\` to the \`internal/server/router.go\` prefix list so the Lambda versioned-API endpoints are isolated from the S3 \`/{bucket}/{key...}\` catch-all (otherwise registering them under the bare prefix panics with a Go ServeMux pattern conflict).

### InvokeEndpoint-less stub-mode

Previously \`Invoke\` errored with \`InvalidParameterValueException: InvokeEndpoint is not configured for this function\` whenever the function was created without kumo's \`invoke_endpoint\` extension — which is always the case for terraform / pulumi / CDK because none of those providers expose the field. Treats the InvokeEndpoint-less case as a stub: accept the invocation, do not run any code, return a benign empty success. RequestResponse returns 200 + \`{}\` Payload, Event returns 202, DryRun returns 204. Functions that DO set \`InvokeEndpoint\` still get the real proxied invocation behaviour.

## Test plan

- [x] \`go test ./internal/service/lambda/... ./internal/server/...\` — green.
- [x] \`make lint\` — clean.
- [x] End-to-end: tofu \`apply\` + \`destroy\` of \`aws_lambda_function { runtime = \"python3.12\"; ... filename = data.archive_file.fn.output_path }\` against \`kumo serve\` — both succeed; before this PR \`apply\` errored sequentially on \`GetFunction\` (path-extract), \`ListVersionsByFunction\` (404), and \`GetFunctionCodeSigningConfig\` (route conflict).
- [x] \`aws lambda invoke --function-name <name> --payload '{}' /tmp/out\` — returns 200, body \`{}\`. \`--invocation-type Event\` returns 202.

Cross-ref: #416, #589.


sivchari (approved):
LGTM

PR #602, opened 2026-05-10
feat(apigateway): bare-prefix routes + DeleteMethod + path-extract fix

mizchi:
## Summary

**Required for terraform compatibility.** terraform-provider-aws sends API Gateway REST API requests to the bare \`/restapis/...\` URL (no \`/apigateway/\` prefix). The routes were registered only under the \`/apigateway/\` prefix, so for the bare-prefix variant the request fell through to the S3 \`/{bucket}/{key...}\` catch-all and returned a \`NoSuchBucket\` XML error — the AWS SDK then panicked trying to parse \`\"invalid character '<'\"\` as JSON.

Found while running tofu against API Gateway for #589.

## Implementation

- **Dual-prefix route registration**: every API Gateway route registration is wrapped in a \`for _, prefix := range []string{\"/apigateway\", \"\"}\` loop so both \`/apigateway/restapis/...\` and \`/restapis/...\` resolve to the same handler. Same pattern as the existing Lambda dual-prefix wrap.
- **Router prefix list**: \`/restapis\` is added to the \`internal/server/router.go\` prefix list so the new bare-prefix routes are isolated from the S3 catch-all. Without this the bare-prefix registration panics with a Go ServeMux pattern conflict (same fix as #601's \`/2019-09-25\` / \`/2020-06-30\` additions).
- **Path-extract helpers tolerate both prefixes**: \`extractResourceParams\` / \`extractRestAPIAndResourceID\` / \`extractMethodParams\` / \`extractIntegrationParams\` / \`extractDeploymentRestAPIID\` / \`extractRestAPIAndDeploymentID\` / \`extractStageRestAPIID\` / \`extractRestAPIAndStageName\` now scan for the \`restapis\` segment anywhere in the path via a new \`pathSegmentsAfterRestapis\` helper. Without this every path-extract returned empty for the bare-prefix variant and storage errored on the empty IDs. Same fix in \`GetResources\` (was inlined).
- **DeleteMethod**: implemented on the service + storage. terraform's \`aws_api_gateway_method\` destroy issues \`DELETE /restapis/{id}/resources/{id}/methods/{method}\` and the route was missing entirely (405 Method Not Allowed), so destroy of any \`aws_api_gateway_rest_api\` that had any method leaked.

## Test plan

- [x] \`go test ./internal/service/apigateway/... ./internal/server/...\` — green.
- [x] \`make lint\` — clean.
- [x] End-to-end against \`kumo serve\`:
  - \`aws_api_gateway_rest_api\` apply + destroy — both succeed; before this PR \`apply\` errored with \`invalid character '<'\` immediately.
  - \`aws_api_gateway_rest_api\` + \`aws_api_gateway_resource\` + \`aws_api_gateway_method\` apply + destroy — all 3 added then all 3 destroyed; before this PR destroy errored on \`DeleteMethod\` (405).

Cross-ref: #416, #589.


sivchari (approved):
LGTM

sivchari:
Approved, but this PR now has merge conflicts (likely from router.go changes). Could you rebase on main? Will merge once conflicts are resolved.

PR #603, opened 2026-05-10
feat(cognito): populate UserPool refresh fields + GetUserPoolMfaConfig stub

mizchi:
## Summary

**Required for terraform compatibility.** terraform-provider-aws's \`resourceUserPoolRead\` dereferences several nested struct pointers on the \`DescribeUserPool\` response without nil-checks (e.g. \`userPool.Policies.PasswordPolicy\` at v5.100.0/internal/service/cognitoidp/user_pool.go:959). Before this PR every \`aws_cognito_user_pool\` apply ended in \`Plugin did not respond\` because the provider segfaulted on the first nil dereference; both \`CreateUserPool\` and \`DescribeUserPool\` had returned 200, but the response was missing the fields the flatten helpers expected.

After fixing the panic the apply gets one step further and then errors on \`InvalidAction: The action GetUserPoolMfaConfig is not valid\` — the post-create refresh also reads MFA config — so this PR also adds \`GetUserPoolMfaConfig\` / \`SetUserPoolMfaConfig\` stubs returning \`OFF\`.

Found while running tofu against Cognito for #589.

## Implementation

- **\`UserPoolOutput\`** grows the fields the provider always reads:
  - \`Arn\` (derived from the region prefix in the user pool ID, e.g. \`us-east-1_AbcdefGhi\` → \`arn:aws:cognito-idp:us-east-1:000000000000:userpool/...\`)
  - \`AdminCreateUserConfig\` / \`AccountRecoverySetting\` / \`VerificationMessageTemplate\` / \`SchemaAttributes\` / \`UserPoolTags\` / \`EstimatedNumberOfUsers\` / \`DeletionProtection\` / \`UserPoolTier\` / \`AliasAttributes\`
  - \`Policies.PasswordPolicy\` (the proximate cause of the segfault — \`userPool.Policies.PasswordPolicy\` was dereferenced without checking \`Policies\`)
- **Defaults match AWS** for an empty user pool — \`MfaConfiguration=OFF\`, \`DeletionProtection=INACTIVE\`, \`UserPoolTier=ESSENTIALS\`, default password policy (8 chars, all classes required, 7-day temp validity), recovery via \`verified_email\` priority 1, default schema with \`sub\` / \`name\` / \`email\` — so terraform's drift detection stays stable across applies.
- **\`GetUserPoolMfaConfig\` / \`SetUserPoolMfaConfig\`** return the AWS-default MFA-OFF shape; storage is not touched.

## Test plan

- [x] \`go test ./internal/service/cognito/...\` — green.
- [x] \`make lint\` — clean.
- [x] End-to-end: tofu \`apply\` + \`destroy\` of \`aws_cognito_user_pool { name = ... }\` against \`kumo serve\` — both succeed; before this PR \`apply\` segfaulted the provider plugin (\`Plugin did not respond\`) immediately after \`DescribeUserPool\` returned 200.

Cross-ref: #416, #589.


sivchari (approved):
LGTM

PR #604, opened 2026-05-10
fix(server): wire SNS topic SQS subscriptions to actually deliver messages

mizchi:
## Summary

SNS \`Publish\` silently dropped every message destined for an SQS subscriber because the \`SqsPublisher\` field on \`sns.MemoryStorage\` was nil — nothing ever called \`SetSQSPublisher\`. The hook was already in place on the SNS side, but no caller wired it to the SQS service.

Found while running a realistic tofu serverless stack (SNS topic + Lambda + DynamoDB + S3) end-to-end and watching:

\`\`\`
aws sns publish --topic-arn <topic> --message hello   # returns MessageId, looks fine
aws sqs receive-message --queue-url <subscribed queue>   # empty — no messages
\`\`\`

## Implementation

- **SQS service exposes \`Storage()\` + \`BaseURL()\`**: minimal getters so cross-service wiring code can hand messages to the storage layer and translate SQS ARNs to queue URLs.
- **\`internal/server/sns_sqs_wiring.go\`** introduces \`snsToSQSPublisher\` — a thin adapter that implements \`sns.SQSPublisher\` by calling \`sqs.Storage.SendMessage\`. Accepts both queue URL and SQS ARN endpoints (subscribers may send either; terraform sends ARNs, raw API calls often pass URLs).
- **\`server.New()\`** calls \`wireSNStoSQS(registry)\` right after the service registry is populated. Type-asserts the SNS storage to \`*sns.MemoryStorage\` to access \`SetSQSPublisher\`; cleanly noops if either service is missing.

Existing SQS-direct send/receive tests are untouched. After this change, SNS publish to a topic with an SQS subscription delivers the raw message body into the queue (RawMessageDelivery=true semantics — the JSON envelope shape AWS uses by default for \`RawMessageDelivery=false\` is a separate follow-up; both worth doing but this PR keeps the scope to \"messages get delivered at all\").

## Test plan

- [x] \`go test ./internal/server/... ./internal/service/sqs/... ./internal/service/sns/...\` — green.
- [x] \`make lint\` — clean.
- [x] End-to-end: \`aws sns publish\` followed by \`aws sqs receive-message\` returns the published body.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -77,6 +77,9 @@ func New(config Config) *Server {
 		srv.RegisterService(svc)
 	}
 
+	// Cross-service wiring (must happen after all services are registered).
+	wireSNStoSQS(registry)
+
 	// Register unified protocol dispatcher for POST /
 	hasJSONServices := len(jsonDispatcher.handlers) > 0
 	hasQueryServices := len(queryDispatcher.handlers) > 0
--- a/internal/server/sns_sqs_wiring.go
+++ b/internal/server/sns_sqs_wiring.go
@@ -1,0 +1,103 @@
+package server
+
+import (
+	"context"
+	"strings"
+
+	"github.com/sivchari/kumo/internal/service"
+	"github.com/sivchari/kumo/internal/service/sns"
+	"github.com/sivchari/kumo/internal/service/sqs"
+)
+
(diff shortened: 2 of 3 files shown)


sivchari (approved):
LGTM

PR #608, opened 2026-05-10
perf(dynamodb): cache key once for Scan sort + pagination (~1.7x)

mizchi:
## Why

`(*Service).Scan` calls `serializeKey` from inside the `sort.Slice`
comparator (so N log N times) and again per item during the
`ExclusiveStartKey` lookup. Each call allocates a `[]string` and a
`strings.Join` result, which makes `serializeKey` the top non-runtime
hot spot under a high-throughput scan workload, and pushes a lot of
work through the GC.

## What

- Pre-compute the key once per item.
- Sort the `(key, item)` pairs.
- Re-use the same keys for `ExclusiveStartKey` lookup.

Behavior is unchanged: same sort order, same pagination semantics.

## Numbers

Workload: 500k `BatchWriteItem` puts followed by a single full `Scan`,
in-memory storage. Apple M2, Go 1.26.

| | Scan + aggregate |
|---|---|
| before | 3.10s |
| after  | 1.82s (~1.7x) |

CPU profile (cum%):
- `serializeKey`: 12.34% → <0.1%
- `gcDrain`: 24.13% → ~25% (similar fraction of a smaller total — fewer allocations means GC has less to scan)

## Test

`go test ./internal/service/dynamodb/...` passes.

### DIFF
--- a/internal/service/dynamodb/storage.go
+++ b/internal/service/dynamodb/storage.go
@@ -693,28 +693,39 @@ func (m *MemoryStorage) Scan(_ context.Context, tableName, filterExpr string, ex
 		results = append(results, m.copyItem(item))
 	}
 
-	// Sort by key for consistent pagination.
-	sort.Slice(results, func(i, j int) bool {
-		keyI := m.serializeKey(td.Table, results[i])
-		keyJ := m.serializeKey(td.Table, results[j])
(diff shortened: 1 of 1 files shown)


sivchari (approved):
LGTM

PR #609, opened 2026-05-10
feat(server): KUMO_PPROF and KUMO_LOG_LEVEL env knobs

mizchi:
## Why

Two small operator knobs that I needed while benchmarking kumo
(1brc-style DynamoDB / S3 throughput tests) and that have general
debugging value.

### KUMO_PPROF / KUMO_PPROF_ADDR

Set \`KUMO_PPROF=1\` to start a separate http server exposing the
standard \`/debug/pprof/{,cmdline,profile,symbol,trace}\` endpoints.
\`KUMO_PPROF_ADDR\` overrides the listen address (default \`:6060\`).

It uses its own \`ServeMux\` so it does not pollute the service router,
and is only started when the env knob is set, so production builds pay
nothing if it stays off.

### KUMO_LOG_LEVEL

Accepts \`debug\`, \`info\`, \`warn\`, \`error\`. Overrides the default INFO.

The router currently logs every request at INFO. That is useful in
development but dominates CPU under throughput tests — in a 30s CPU
profile of a heavy DynamoDB scan workload, \`log/slog\` + \`syscall.write\`
together accounted for ~16% of samples. Being able to set the level
without rebuilding makes profiles representative of handler work
instead of slog overhead.

## Test

\`go test ./internal/server/...\` passes.

### DIFF
--- a/internal/server/pprof.go
+++ b/internal/server/pprof.go
@@ -22,7 +22,9 @@ func startPprofServer(logger *slog.Logger) {
 	if v == "" {
 		return
 	}
+
 	on, err := strconv.ParseBool(v)
+
 	if err != nil || !on {
 		return
 	}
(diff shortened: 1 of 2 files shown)


sivchari (approved):
LGTM

sivchari:
Approved, but this PR now has merge conflicts from #612 merge (both modify DefaultConfig). Could you rebase on main? Will merge once conflicts are resolved.

PR #612, opened 2026-05-10
fix(server): honor KUMO_HOST/KUMO_PORT and --host/--port

izumin5210:
## Why

The README documents `--host` / `--port` and `KUMO_HOST` / `KUMO_PORT`,
but `kumo` always listened on `0.0.0.0:4566` — env vars were never read
and the flags never reached `Config`. `kumo serve --port ...` also
failed as an unknown flag.

## What

- `DefaultConfig` reads `KUMO_HOST` / `KUMO_PORT` (unparseable port
  keeps the default).
- `--host` / `--port` are wired into `Config` and registered on each
  server-starting command (root, `serve`), not on
  `root.PersistentFlags`, so client subcommands like `s3` / `acm`
  do not inherit a server-only flag.
- `--port` is an `Int` flag so `pflag` rejects invalid values
  automatically.

### DIFF
--- a/cmd/kumo/main.go
+++ b/cmd/kumo/main.go
@@ -93,8 +93,17 @@ func main() {
 
 	// Root command starts the server when no CLI subcommand is matched.
 	// Docker uses `kumo --host 0.0.0.0 --port 4566`, so we accept these flags.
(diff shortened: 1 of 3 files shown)


sivchari (approved):
LGTM

PR #617, opened 2026-05-11
fix(s3): decode URL-encoded x-amz-copy-source before parsing

### LINKED ISSUE #616: S3 CopyObject rejects URL-encoded x-amz-copy-source header value
## Summary

  `s3.CopyObject` returns `InvalidArgument: Invalid copy source` when
  `x-amz-copy-source` is sent in URL-encoded form (e.g. `bucket%2Fkey`),
  even though AWS S3 accepts both URL-encoded and plain forms. The
  existing comment in `parseCopySource` ("Format: /bucket/key or
  bucket/key (URL-encoded)") suggests URL-encoded is intended to be
  supported, but the implementation splits on a literal `/` without
  URL-decoding first.

  ## Environment

  - kumo: v0.19.0
  - Client: AWS SDK for Go v2 (`github.com/aws/aws-sdk-go-v2/service/s3`)

  ## Reproduction

  ```go
  client := s3.NewFromConfig(cfg, func(o *s3.Options) {
      o.BaseEndpoint = aws.String("http://localhost:4566")
      o.UsePathStyle = true
  })

  _, _ = client.CreateBucket(ctx, &s3.CreateBucketInput{Bucket: aws.String("b")})
  _, _ = client.PutObject(ctx, &s3.PutObjectInput{
      Bucket: aws.String("b"), Key: aws.String("src.txt"),
      Body: strings.NewReader("hi"),
  })

  // Caller URL-encodes the CopySource (a pattern used by some apps to
  // safely handle keys containing '/', '+', spaces, etc.):
  copySource := url.PathEscape("b/src.txt") // "b%2Fsrc.txt"
  _, err := client.CopyObject(ctx, &s3.CopyObjectInput{
      Bucket: aws.String("b"), Key: aws.String("dst.txt"),
      CopySource: aws.String(copySource),
  })
  // err: InvalidArgument: Invalid copy source

  Same request against real AWS S3 succeeds.

  Root Cause

  internal/service/s3/handlers.go:647-658:

  func parseCopySource(source string) (bucket, key string) {
      source = strings.TrimPrefix(source, "/")
      idx := strings.IndexByte(source, '/')   // looks for literal '/'
      if idx < 0 {
          return "", ""                       // URL-encoded form falls through here
      }
      return source[:idx], source[idx+1:]
  }

  When the entire path is URL-encoded (b%2Fsrc.txt), there is no
  literal / byte, so IndexByte returns -1 and the function returns
  empty strings.

  Proposed Fix

  URL-decode the source before splitting. url.PathUnescape is the
  right choice (path-style, preserves +).

  import "net/url"

  func parseCopySource(source string) (bucket, key string) {
      // AWS allows x-amz-copy-source to be URL-encoded. Decode first so
      // both "bucket/key" and "bucket%2Fkey" produce the same result.
      if decoded, err := url.PathUnescape(source); err == nil {
          source = decoded
      }
      source = strings.TrimPrefix(source, "/")
      idx := strings.IndexByte(source, '/')
      if idx < 0 {
          return "", ""
      }
      return source[:idx], source[idx+1:]
  }

  The same fix should be applied where parseCopySource is reused in
  UploadPartCopy (line ~1738).

  Why this matters

  Many production AWS Go applications URL-encode CopySource to be safe
  with keys containing /, +, or spaces. These apps work against real
  AWS S3 but fail against kumo, making kumo a leaky replacement for
  local-vs-prod parity testing.

hc100:
## Summary

Fixes #616: `S3.CopyObject` rejected requests whose `x-amz-copy-source` header was sent in URL-encoded form (`bucket%2Fkey`), even though AWS S3 itself accepts both URL-encoded and plain forms. The existing comment on `parseCopySource` already documented the intent ("Format: /bucket/key or bucket/key (URL-encoded)"), but the implementation split on a literal `/` without decoding first, so the encoded form fell through to the "no separator" branch and returned empty strings — surfacing as `InvalidArgument: Invalid copy source`.

## Fix

URL-decode the header value with `url.PathUnescape` before splitting. `PathUnescape` (not `QueryUnescape`) is used so that `+` is preserved as a valid S3 key character.

```go
if decoded, err := url.PathUnescape(source); err == nil {
    source = decoded
}
source = strings.TrimPrefix(source, "/")
idx := strings.IndexByte(source, '/')
```

The same `parseCopySource` is reused by `UploadPartCopy`, so the fix applies there transparently.

## Behavior matrix

| Input | Before | After |
|---|---|---|
| `bucket/key.txt` | ✅ ok | ✅ ok |
| `/bucket/key.txt` | ✅ ok | ✅ ok |
| `bucket%2Fkey.txt` | ❌ `InvalidArgument` | ✅ ok |
| `bucket/path%2Fto%2Fkey.txt` | ❌ key returned as `path%2Fto%2Fkey.txt` | ✅ decoded to `path/to/key.txt` |
| `bucket/file+name.txt` | ✅ ok | ✅ `+` preserved |

## Changes

- `internal/service/s3/handlers.go`: decode source before splitting in `parseCopySource`.
- `internal/service/s3/copysource_test.go` (new): table-driven unit tests covering plain / leading-slash / encoded-separator / encoded-subpath / fully-encoded / plus-preserved / no-separator / empty.
- `test/integration/s3_test.go`: added `TestS3_CopyObject_URLEncodedSource` which builds CopySource via `url.PathEscape` (the form the AWS Go SDK callers in the wild produce) and round-trips through `PutObject` → `CopyObject` → `GetObject`.

## Test plan

- [x] `make test` — unit tests pass
- [x] `go test ./internal/service/s3/...` — all S3 unit tests pass including new `TestParseCopySource` (8 sub-cases)
- [x] `go vet -tags=integration ./test/integration/...` — clean
- [ ] `make test-integration` — please run in CI (requires Docker)
- [ ] `make lint` — please run in CI (golangci-lint not installed locally)

### DIFF
--- a/internal/service/s3/copysource_test.go
+++ b/internal/service/s3/copysource_test.go
@@ -1,0 +1,38 @@
+package s3
+
+import "testing"
+
+// TestParseCopySource covers the shapes AWS clients send in the
+// x-amz-copy-source header: plain, leading-slash, and URL-encoded.
+// AWS S3 accepts all of these so kumo must too.
+func TestParseCopySource(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		name       string
+		source     string
+		wantBucket string
+		wantKey    string
+	}{
+		{"plain", "bucket/key.txt", "bucket", "key.txt"},
+		{"leading slash", "/bucket/key.txt", "bucket", "key.txt"},
+		{"encoded separator", "bucket%2Fkey.txt", "bucket", "key.txt"},
+		{"encoded subpath", "bucket/path%2Fto%2Fkey.txt", "bucket", "path/to/key.txt"},
+		{"fully encoded with leading slash", "%2Fbucket%2Fkey.txt", "bucket", "key.txt"},
+		{"plus preserved", "bucket/file+name.txt", "bucket", "file+name.txt"},
+		{"no separator", "bucket", "", ""},
+		{"empty", "", "", ""},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			gotBucket, gotKey := parseCopySource(tc.source)
+			if gotBucket != tc.wantBucket || gotKey != tc.wantKey {
+				t.Errorf("parseCopySource(%q) = (%q, %q), want (%q, %q)",
+					tc.source, gotBucket, gotKey, tc.wantBucket, tc.wantKey)
+			}
+		})
+	}
+}
--- a/internal/service/s3/handlers.go
+++ b/internal/service/s3/handlers.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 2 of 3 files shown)


sivchari (approved):
Nice, thx!

PR #621, opened 2026-05-11
Add ChangeMessageVisibilityRequest struct

MahmudovMZ:
Added ChangeMessageVisibilityRequest struct for SQS.

### DIFF
--- a/internal/service/sqs/types.go
+++ b/internal/service/sqs/types.go
@@ -251,6 +251,13 @@ type SetQueueAttributesRequest struct {
 	Attributes map[string]string `json:"Attributes"`
 }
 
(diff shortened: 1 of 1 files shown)


sivchari:
Thanks for the contribution! The storage implementation alone isn't enough for this to work end-to-end -- please check `CONTRIBUTING.md` for the full checklist when adding a new operation. You'll also need handler wiring and integration tests.

Also, could you merge this into #622? Having the types and implementation in separate PRs makes it hard to review as a unit.

sivchari:
Updating my earlier comment -- please fold this into #622 along with the handler wiring and integration test, then close this PR. That way everything for ChangeMessageVisibility lands in one place.

PR #618, opened 2026-05-11
Add ChangeMessageVisibility method to MemoryStorage

MahmudovMZ:
Implemented ChangeMessageVisibility method to update message visibility timeout.

PR #620, opened 2026-05-11
Add ChangeMessageVisibilityRequest struct

MahmudovMZ:
Added ChangeMessageVisibilityRequest struct for SQS.

PR #622, opened 2026-05-11
Implement ChangeMessageVisibility method

MahmudovMZ:
Added ChangeMessageVisibility method to modify the visibility timeout of inflight messages.


sivchari:
Thanks! The storage layer looks good, but this isn't wired up yet -- requests will hit `InvalidAction` because `DispatchAction` in `handlers.go` doesn't have a `ChangeMessageVisibility` case. Please check `CONTRIBUTING.md` for the full checklist.

What's missing:
- Handler method in `handlers.go` that parses the request and calls the storage method
- `case "ChangeMessageVisibility"` in `DispatchAction`
- Integration test

Also, could you fold #621 into this PR so everything lands together?

sivchari:
Could you consolidate all ChangeMessageVisibility work into this single PR? That means:

1. The type definition from #621 (`ChangeMessageVisibilityRequest` in `types.go`)
2. The storage implementation (already here)
3. Handler method in `handlers.go` that parses the request JSON and calls the storage method
4. `case "ChangeMessageVisibility"` in `DispatchAction`
5. Integration test in `test/integration/sqs_test.go`

Please check `CONTRIBUTING.md` for the full checklist. Once this PR has everything, close #618, #620, and #621.

MahmudovMZ:
Got it! I'll add the type definition, handler method, DispatchAction case, and integration test to this PR, then close #618, #620, and #621.

MahmudovMZ:
All changes consolidated into this PR: type definition, handler, DispatchAction case, and integration test added. Ready for review

sivchari (changes requested):
Thanks for the implementation! The handler and storage logic look solid.

One thing to fix before merging: the integration test should use the golden test pattern with `golden.New(t, ...).Assert(...)` instead of direct assertions. This is the standard pattern used across all kumo integration tests (see `TestSQS_QueuePolicy` or `TestSQS_CreateAndDeleteQueue` for examples).

For `ChangeMessageVisibility` which returns an empty body, you can still assert the response shape:

```go
golden.New(t, golden.WithIgnoreFields("ResultMetadata")).Assert(t.Name()+"_change_visibility", changeOutput)
```

Please update the test to use golden files, and add the generated `.golden.go` files to the commit.

MahmudovMZ:
Got it, will update the integration test to use the golden test pattern. On it!

MahmudovMZ:
The failing tests (TestRoute53Resolver_ResolverEndpoint, TestS3_EventBridgeNotification) are unrelated to this PR - they appear to have pre-existing golden file mismatches due to dynamic values like ARNs and timestamps. My SQS changes pass successfully

MahmudovMZ:
I've updated the golden test pattern for ChangeMessageVisibility as requested. The other failing tests appear to be pre-existing issues — they fail with 500 errors from LocalStack for operations that aren't supported, or have outdated golden files unrelated to my changes. My SQS implementation and its tests pass successfully.

PR #623, opened 2026-05-12
fix(dynamodb): cli accepts object form for --local/global-secondary-indexes

hirasawayuki:
## Summary
`kumo dynamodb create-table --local-secondary-indexes` (and `--global-secondary-indexes`) silently produced a table with no secondary indexes when the value was a single JSON object instead of an array. The same JSON works against the real AWS DynamoDB API via `aws` CLI, so kumo should accept the same input for parity.

## Problem

`parseLSI` / `parseGSI` only unmarshal into `[]struct{...}` and discard the result of `json.Unmarshal` via `_ = json.Unmarshal(...)`. Passing a single object (`{...}`) fails to unmarshal, the error is dropped, and the resulting empty slice is treated as "no indexes specified".

```bash
docker exec kumo kumo dynamodb create-table \
  --table-name lsi-via-kumo \
  --attribute-definitions AttributeName=PK,AttributeType=S \
  --attribute-definitions AttributeName=SK,AttributeType=S \
  --attribute-definitions AttributeName=Idx,AttributeType=N \
  --key-schema AttributeName=PK,KeyType=HASH \
  --key-schema AttributeName=SK,KeyType=RANGE \
  --local-secondary-indexes '{"IndexName":"IdxName","KeySchema":[...],"Projection":{"ProjectionType":"ALL"}}' \
  --provisioned-throughput ReadCapacityUnits=1,WriteCapacityUnits=1

# describe-table afterwards:
#   .Table.LocalSecondaryIndexes == null   ❌
  ```

The same payload via `aws --endpoint-url http://localhost:4566 dynamodb create-table ...` works, because the `aws` CLI wraps a bare object into an array client-side before sending the DynamoDB API request. The kumo HTTP API already accepts the array form, so applying the same normalization in the CLI lines the two up.

## Fix

In `cli/dynamodb.go`:

- Add `normalizeIndexJSON` which wraps `{...}` into `[{...}]` (mirrors the `aws` CLI normalization).
- Change `parseLSI` / `parseGSI` to return `(values, error)` so a JSON parse failure is no longer swallowed.
- Propagate the error from `newDynamoDBCreateTableCmd` so the command exits non-zero instead of exiting 0 on failure.

  ## Test plan

  - [x] `go build ./...`
  - [x] `make lint` → 0 issues
  - [x] Manually exercised against a running kumo container at `http://localhost:4566`:
    - `--local-secondary-indexes '{...}'` (object) → `describe-table` shows the LSI ✅
    - `--local-secondary-indexes '[{...}]'` (array) → still works, no regression ✅
    - `--local-secondary-indexes '{ broken json'` → `failed to parse --local-secondary-indexes: invalid json: ...` and exit 1 ✅

### DIFF
--- a/cli/dynamodb.go
+++ b/cli/dynamodb.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"unicode"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
@@ -67,11 +68,21 @@ func newDynamoDBCreateTableCmd() *cobra.Command {
 			}
 
 			if gsiJSON != "" {
-				input.GlobalSecondaryIndexes = parseGSI(gsiJSON)
+				gsis, err := parseGSI(gsiJSON)
+				if err != nil {
+					return err
+				}
+
+				input.GlobalSecondaryIndexes = gsis
 			}
 
 			if lsiJSON != "" {
-				input.LocalSecondaryIndexes = parseLSI(lsiJSON)
+				lsis, err := parseLSI(lsiJSON)
+				if err != nil {
+					return err
+				}
+
+				input.LocalSecondaryIndexes = lsis
 			}
 
 			out, err := client.CreateTable(cmd.Context(), input)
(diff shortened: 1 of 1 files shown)


sivchari (approved):
Thx!

PR #669, opened 2026-05-13
Add fuzz coverage for S3 and DynamoDB edge cases

mizchi:
## Summary

Adds targeted Go fuzz coverage for S3 and DynamoDB edge cases, then fixes the issues found by the new fuzz seeds.

## Bugs Found

- S3 byte range parsing accepted suffix ranges for zero-length objects. For example, `Range: bytes=-1` with a zero-length object produced an invalid successful range `(0, -1)`, which could lead to an invalid `206 Partial Content` response such as `Content-Range: bytes 0--1/0`.
- DynamoDB condition evaluation could panic when `contains()` was evaluated against a list attribute containing a JSON `null` element, because the list walker dereferenced a nil `*AttributeValue`.

## Fuzz Coverage Added

- `FuzzParseByteRangeInvariants`
  - checks S3 byte range parser bounds and zero-length object behavior
- `FuzzParseCopySourceAndRangeNoPanic`
  - exercises S3 copy source and upload-part-copy range parsing
- `FuzzConditionExpressionNoPanic`
  - exercises DynamoDB condition expressions against fuzzed item/value JSON
- `FuzzAttributeValueJSONRoundTrip`
  - exercises DynamoDB `AttributeValue` JSON unmarshal/marshal/unmarshal stability

## Validation

- `go test ./internal/service/s3 -run '^$' -fuzz=FuzzParseByteRangeInvariants -fuzztime=15s -parallel=4`
- `go test ./internal/service/s3 -run '^$' -fuzz=FuzzParseCopySourceAndRangeNoPanic -fuzztime=15s -parallel=4`
- `go test ./internal/service/dynamodb -run '^$' -fuzz=FuzzConditionExpressionNoPanic -fuzztime=15s -parallel=4`
- `go test ./internal/service/dynamodb -run '^$' -fuzz=FuzzAttributeValueJSONRoundTrip -fuzztime=15s -parallel=4`
- `go test ./internal/service/s3 ./internal/service/dynamodb -count=1 -v`
- `make test`
- `make lint`
- `git diff --check`

PR #674, opened 2026-05-13
Fix DynamoDB UpdateExpression invalid UTF-8 panic

### LINKED ISSUE #671: DynamoDB UpdateExpression parser panics on invalid UTF-8
Discovered while extending fuzz coverage in #669.

## Problem
An invalid UTF-8 update expression such as `\x98 REMOVE` can panic in `parseUpdateClauses`. The parser uppercases the expression with `strings.ToUpper`, then uses indexes from the uppercased string to slice the original string. Invalid UTF-8 can change the byte length and make those indexes unsafe.

## Impact
A request-controlled `UpdateExpression` can crash the DynamoDB handler path.

## Expected behavior
Invalid or unusual byte sequences in an update expression should not panic the server.

## Regression coverage
The follow-up PR should include a deterministic regression test that exercises the invalid UTF-8 expression.

mizchi:
## Summary
- avoid byte-index drift in `parseUpdateClauses` by uppercasing ASCII bytes without changing length
- add regression coverage for an invalid UTF-8 update expression that previously panicked

Fixes #671.
Related to #669.

## Test
- `go test ./internal/service/dynamodb -run TestUpdateItemInvalidUTF8ExpressionDoesNotPanic -count=1`
- `go test ./internal/service/dynamodb -count=1`
- `make lint`
- `git diff --check`

### DIFF
--- a/internal/service/dynamodb/storage.go
+++ b/internal/service/dynamodb/storage.go
@@ -998,7 +998,7 @@ type updateClause struct {
 // parseUpdateClauses splits an update expression into individual clauses.
 func parseUpdateClauses(expr string) []updateClause {
 	keywords := []string{"SET", "ADD", "DELETE", "REMOVE"}
-	upper := strings.ToUpper(expr)
+	upper := asciiUpper(expr)
 
 	type pos struct {
 		idx    int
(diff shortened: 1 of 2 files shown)


sivchari (approved):
LGTM. The asciiUpper fix correctly prevents the UTF-8 index mismatch panic in parseUpdateClauses.

PR #675, opened 2026-05-13
Reject empty S3 multipart completion

### LINKED ISSUE #672: S3 CompleteMultipartUpload accepts an empty part list
Discovered while extending fuzz coverage in #669.

## Problem
`CompleteMultipartUpload` currently accepts an empty `Part` list, returns success, creates an empty object, and removes the multipart upload state.

## Impact
This is incompatible with S3 behavior and can lose the in-progress multipart upload state for an invalid completion request.

## Expected behavior
Completing a multipart upload with no parts should return `InvalidArgument` and keep the upload available.

## Regression coverage
The follow-up PR should include a deterministic regression test for the empty part list case.

mizchi:
## Summary
- reject `CompleteMultipartUpload` requests with an empty part list
- preserve the in-progress multipart upload after the invalid completion attempt
- add regression coverage for the empty part list case

Fixes #672.
Related to #669.

## Test
- `go test ./internal/service/s3 -run TestCompleteMultipartUploadRejectsEmptyPartList -count=1`
- `go test ./internal/service/s3 -count=1`
- `make lint`
- `git diff --check`

### DIFF
--- a/internal/service/s3/blob_test.go
+++ b/internal/service/s3/blob_test.go
@@ -114,6 +114,8 @@ func TestMemoryStorage_VersionedBodiesRoundTrip(t *testing.T) {
 }
 
 func TestMemoryStorage_LegacyInlineBodyMigrates(t *testing.T) {
+	const legacyBody = "hello"
+
 	ctx := context.Background()
 	dir := t.TempDir()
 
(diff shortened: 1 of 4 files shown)


sivchari (changes requested):
The fix correctly rejects empty part lists and preserves the upload for retry.

One issue: the error code is `InvalidArgument`, but AWS S3 returns `MalformedXML` (HTTP 400) for an empty parts list in `CompleteMultipartUpload`. Please change the error code to `MalformedXML`.

PR #679, opened 2026-05-13
fix(dynamodb): validate key schema and write actions

### LINKED ISSUE #676: DynamoDB accepts invalid key schema and empty write actions
Found while extending the fuzz/regression work from #669.

## Problem
DynamoDB storage currently serializes keys without validating them against the table key schema. This lets requests succeed when the hash/range key is missing or uses an unsupported type, and it can create inaccessible or colliding in-memory items.

The same validation gap also appears in batch/transaction writes: empty WriteRequest / TransactWriteItem members are accepted as no-ops instead of returning a validation error.

## Expected
- Key maps must match the table key schema.
- Put-like item writes must include valid key attributes.
- Batch/transaction write members must contain exactly one action.

## Regression coverage
Add storage-level regression tests for missing keys, wrong key types, incomplete composite keys, empty batch writes, empty transaction writes, and batch get key validation.

mizchi:
## Summary
- validate DynamoDB primary key maps against the table key schema before serializing them
- reject missing/wrong key attributes for Put/Get/Delete/Update, BatchGet, BatchWrite, TransactGet, and TransactWrite paths
- reject empty or multi-action BatchWriteItem / TransactWriteItems members instead of treating them as no-ops

Fixes #676.
Refs #669.

## Tests
- go test ./internal/service/dynamodb -count=1
- make lint
- git diff --check

### DIFF
--- a/internal/service/dynamodb/storage.go
+++ b/internal/service/dynamodb/storage.go
@@ -401,6 +401,10 @@ func (m *MemoryStorage) PutItem(_ context.Context, tableName string, item Item, 
 		}
 	}
 
+	if err := validateItemKey(td.Table, item); err != nil {
+		return nil, err
+	}
+
 	key := m.serializeKey(td.Table, item)
 
 	// Evaluate condition against existing item (nil if not exists).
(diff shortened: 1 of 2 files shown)


sivchari (approved):
LGTM

sivchari:
Approved, but rebase merge is failing due to recent changes on main. Could you rebase on latest main? Ready to merge once resolved.

PR #681, opened 2026-05-13
fix(range): reject suffix ranges on empty bodies

### LINKED ISSUE #678: S3 and CloudFront accept suffix ranges on empty bodies
Found while extending the fuzz/regression work from #669.

## Problem
The S3 and CloudFront byte range parsers treat a suffix range against an empty body, such as bytes=-1 with total size 0, as satisfiable and return start=0/end=-1. This can produce invalid 206-style range metadata for empty objects or cached bodies.

## Expected
Suffix ranges for empty bodies should be unsatisfiable and fall through to the existing InvalidRange / full-fetch behavior.

## Regression coverage
Add parser-level regression tests for empty-body suffix ranges in both S3 and CloudFront cache.

mizchi:
## Summary
- reject suffix byte ranges when total size is zero in S3 range parsing
- apply the same empty-body suffix guard to CloudFront cache range parsing
- add parser regression tests for both services

Fixes #678.
Refs #669.

## Tests
- go test ./internal/service/s3 ./internal/service/cloudfront/cache -count=1
- make lint
- git diff --check

### DIFF
--- a/internal/service/cloudfront/cache/rules.go
+++ b/internal/service/cloudfront/cache/rules.go
@@ -586,7 +586,7 @@ func splitByteRangeSpec(header string) (string, string, bool) {
 // parseSuffixRange resolves `bytes=-N` (the last N bytes).
 func parseSuffixRange(endRaw string, totalSize int64) (int64, int64, bool) {
 	n, err := strconv.ParseInt(endRaw, 10, 64)
-	if err != nil || n <= 0 {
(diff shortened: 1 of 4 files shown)


sivchari (approved):
LGTM. Correct guard for suffix ranges on empty bodies, applied consistently to both S3 and CloudFront cache.

PR #683, opened 2026-05-13
fix(messaging): validate SNS and SQS names

### LINKED ISSUE #682: SNS and SQS accept delimiter characters in resource names
Found while extending the regression work from #669.

## Problem
SNS topic names and SQS queue names are accepted without validating the character set. Names containing ARN or URL delimiters such as colon, slash, or percent-encoded slash can create malformed resource identifiers and ambiguous reverse lookup behavior.

For SNS, a colon in the topic name also corrupts subscription ARN construction because the topic name is recovered by splitting the topic ARN on colons.

## Expected
SNS and SQS should reject names outside the AWS-compatible resource name character set before constructing ARNs or queue URLs.

## Regression coverage
Add storage-level regression tests for delimiter names and keep valid standard/FIFO names accepted.

mizchi:
## Summary
- reject SNS topic names that contain ARN/path delimiter characters or unsupported dots
- reject SQS queue names that contain URL/ARN delimiter characters or unsupported dots
- keep valid standard and FIFO-style names accepted
- add storage-level regression tests for the delimiter cases

Fixes #682.
Refs #669.

## Tests
- go test ./internal/service/sns ./internal/service/sqs -count=1
- make lint
- git diff --check

### DIFF
--- a/internal/awsname/messaging.go
+++ b/internal/awsname/messaging.go
@@ -1,0 +1,39 @@
+// Package awsname holds resource-name validators shared between AWS service
+// emulators (currently SNS topics, SQS queues — same character rules, different
+// length caps).
+package awsname
+
+import "strings"
+
+// IsValidMessagingResource returns true when name satisfies the character /
+// suffix rules AWS applies to SNS topic and SQS queue names:
+//
+//   - 1..maxLen characters
+//   - alphanumeric, underscore, hyphen
+//   - a single trailing ".fifo" is permitted (FIFO queue / topic), and the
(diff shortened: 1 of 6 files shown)


sivchari (changes requested):
The validation rules are correct for both SNS and SQS.

However, `isValidMessagingName` in SNS and `isValidQueueName` in SQS implement essentially the same character validation logic independently. Please have `isValidQueueName` delegate to `isValidMessagingName(name, 80)` (or extract a shared helper) to avoid the duplication.

PR #685, opened 2026-05-13
fix(kinesis): validate record key inputs

### LINKED ISSUE #684: Kinesis accepts invalid record key values
Found while extending the regression work from #669.

## Problem
Kinesis PutRecord and PutRecords accept invalid record key values.

- A non-numeric ExplicitHashKey is passed to big.Int.SetString without checking whether parsing succeeded, so invalid values can be routed as hash key zero instead of being rejected.
- An empty PartitionKey is accepted and hashed, although Kinesis requires a non-empty partition key.

## Expected
Invalid ExplicitHashKey and empty PartitionKey values should return InvalidArgumentException for PutRecord and entry-level failures for PutRecords.

## Regression coverage
Add storage-level tests for invalid ExplicitHashKey and empty PartitionKey on both PutRecord and PutRecords.

mizchi:
## Summary
- reject non-numeric ExplicitHashKey values for PutRecord
- reject empty PartitionKey values for PutRecord
- return entry-level InvalidArgumentException for invalid ExplicitHashKey / PartitionKey values in PutRecords
- share the Kinesis max hash key constant with range calculation
- add storage-level regression tests for both paths

Fixes #684.
Refs #669.

## Tests
- go test ./internal/service/kinesis -count=1
- make lint
- git diff --check

### DIFF
--- a/internal/service/kinesis/storage.go
+++ b/internal/service/kinesis/storage.go
@@ -30,6 +30,7 @@ const (
 	defaultRetentionHours   = 24
 	maxRecordsPerGet        = 10000
 	shardIteratorExpiration = 5 * time.Minute
+	maxHashKeyString        = "340282366920938463463374607431768211455"
 )
 
 // Storage defines the Kinesis storage interface.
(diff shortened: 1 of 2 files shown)


sivchari (approved):
LGTM

PR #687, opened 2026-05-13
fix(kinesis): validate CreateStream inputs

### LINKED ISSUE #686: Kinesis CreateStream accepts invalid stream inputs
Follow-up from #669.

While probing Kinesis compatibility gaps, I found that CreateStream currently accepts and stores inputs that AWS rejects:

- empty StreamName
- StreamName longer than 128 characters
- StreamName values containing ARN/path delimiters or characters outside [a-zA-Z0-9_.-]
- supplied ShardCount values below 1
- unsupported StreamModeDetails.StreamMode values

This can also create malformed stream ARNs for names containing delimiters, because Kumo builds arn:aws:kinesis:...:stream/<name> and other handlers parse stream names from ARNs by splitting on slash.

Expected behavior: reject these inputs before storing the stream, using an InvalidArgumentException-style 400 response.

Actual behavior: Kumo persists the invalid stream, or silently defaults invalid ShardCount values to one shard.

Reference: AWS CreateStream documents StreamName as required, length 1..128, pattern [a-zA-Z0-9_.-]+; ShardCount minimum 1 when supplied; StreamMode values PROVISIONED or ON_DEMAND.

mizchi:
## Summary
- Fixes #686
- Related to #669
- reject invalid CreateStream StreamName values before storing streams
- reject supplied ShardCount values below 1
- reject unsupported StreamModeDetails.StreamMode values
- keep shard creation separated from request validation so invalid requests cannot persist malformed ARNs

## Tests
- go test ./internal/service/kinesis -count=1
- go test ./... -count=1
- make lint
- git diff --check

### DIFF
--- a/internal/service/kinesis/handlers.go
+++ b/internal/service/kinesis/handlers.go
@@ -448,6 +448,8 @@ func getErrorStatus(code string) int {
 		return http.StatusConflict
 	case errInvalidArgument:
 		return http.StatusBadRequest
+	case errValidation:
+		return http.StatusBadRequest
 	case errExpiredIterator:
 		return http.StatusBadRequest
 	default:
--- a/internal/service/kinesis/storage.go
+++ b/internal/service/kinesis/storage.go
@@ -23,6 +23,7 @@ const (
 	errResourceInUse    = "ResourceInUseException"
 	errInvalidArgument  = "InvalidArgumentException"
 	errExpiredIterator  = "ExpiredIteratorException"
+	errValidation       = "ValidationException"
 )
 
 // Default values.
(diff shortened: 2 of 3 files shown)


sivchari (changes requested):
The validation logic and refactoring are clean.

One issue: the error code returned for all three validations (invalid name, invalid shard count, invalid stream mode) is `InvalidArgumentException`. AWS Kinesis returns `ValidationException` for these cases. SDK clients that match on the exception type will see a different code than what real AWS returns.

Please change the error code to `ValidationException`.

PR #689, opened 2026-05-13
fix(s3): honor CopyObject metadata directive

### LINKED ISSUE #688: S3 CopyObject ignores metadata replacement directive
Follow-up from #669.

CopyObject currently always writes the destination object with the source object metadata. It ignores the x-amz-metadata-directive header, so requests that use REPLACE still keep stale source metadata.

This is visible with a request like:

- source object metadata: color=blue, Content-Type=text/plain
- CopyObject request headers: x-amz-metadata-directive=REPLACE, x-amz-meta-color=red, Content-Type=application/json

Expected behavior: the destination object should use only the metadata provided by the copy request.

Actual behavior: the destination object keeps color=blue and Content-Type=text/plain from the source object.

This matters for SDK and Terraform flows that update S3 object metadata by copying the object onto a new key or onto itself. AWS S3 documents MetadataDirective as COPY or REPLACE, with REPLACE using metadata supplied in the request.

mizchi:
## Summary
- Fixes #688
- Related to #669
- honor x-amz-metadata-directive on CopyObject
- keep COPY as the default by cloning source metadata
- support REPLACE by extracting Content-Type and x-amz-meta-* headers from the copy request
- reject unsupported metadata directive values before storing the destination object

## Tests
- go test ./internal/service/s3 -run TestCopyObject.*Metadata -count=1
- go test ./... -count=1
- make lint
- git diff --check

### DIFF
--- a/internal/service/s3/copyobject_test.go
+++ b/internal/service/s3/copyobject_test.go
@@ -1,0 +1,123 @@
+package s3
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+func TestCopyObjectReplacesMetadataWhenDirectiveIsReplace(t *testing.T) {
+	t.Parallel()
+
+	store, svc := setupCopyObjectMetadataFixture(t)
+	w := issueCopyObject(svc, map[string]string{
+		"X-Amz-Metadata-Directive": "REPLACE",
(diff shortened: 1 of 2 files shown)


sivchari (approved):
LGTM

sivchari:
Approved, but rebase merge is failing due to recent changes on main. Could you rebase on latest main? Ready to merge once resolved.

PR #691, opened 2026-05-13
fix(s3): honor CopyObject tagging directive

### LINKED ISSUE #690: S3 CopyObject does not preserve or replace object tags
Follow-up from #669. Related to #689, but this tracks object tags rather than metadata.

CopyObject currently creates the destination object without copying source object tags and without honoring x-amz-tagging-directive / x-amz-tagging.

Cases that currently diverge from S3:

- default CopyObject should copy source tags to the destination object
- x-amz-tagging-directive=REPLACE should apply tags from the x-amz-tagging request header
- unsupported tagging directive values should be rejected before storing the destination object

This matters for SDK and Terraform flows that copy S3 objects while preserving tags, or replace tags during copy. AWS S3 exposes TaggingDirective as COPY or REPLACE for CopyObject.

mizchi:
## Summary
- Fixes #690
- Related to #669 and #689
- copy source object tags by default during CopyObject
- support x-amz-tagging-directive=REPLACE with tags from x-amz-tagging
- reject unsupported tagging directive values before storing the destination object

## Tests
- go test ./internal/service/s3 -run TestCopyObject.*Tags|TestCopyObjectRejectsInvalidTaggingDirective -count=1
- go test ./... -count=1
- make lint
- git diff --check

### DIFF
--- a/internal/service/s3/copyobject_tagging_test.go
+++ b/internal/service/s3/copyobject_tagging_test.go
@@ -1,0 +1,115 @@
+package s3
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+func TestCopyObjectCopiesSourceTagsByDefault(t *testing.T) {
+	t.Parallel()
+
+	store, svc := setupCopyObjectTaggingFixture(t)
+	w := issueTaggedCopyObject(svc, nil)
+
+	if w.Code != http.StatusOK {
+		t.Fatalf("CopyObject status: got %d, want %d (body=%s)", w.Code, http.StatusOK, w.Body.String())
+	}
+
+	tags, err := store.GetObjectTagging(context.Background(), "dst", "copied.txt")
+	if err != nil {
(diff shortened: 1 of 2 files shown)


sivchari (changes requested):
The tagging directive logic itself is correct and matches the AWS spec.

One issue: this PR introduces `parseObjectTaggingHeader` which uses `url.ParseQuery` to parse `x-amz-tagging`, but the existing `parseTaggingHeader` (used by `PutObject`) does manual `strings.Cut` without URL decoding. This means percent-encoded tag keys/values (e.g. `key%20name=val`) will be handled differently between `PutObject` and `CopyObject` with `REPLACE` directive.

Please unify the two parsers — either replace the existing `parseTaggingHeader` with the `url.ParseQuery`-based version, or have `parseObjectTaggingHeader` delegate to the existing one.

sivchari (approved):
LGTM

sivchari:
LGTM on the parser unification. Rebase merge is failing because main has diverged. Could you rebase on latest main?

PR #693, opened 2026-05-13
fix(s3): honor copy source version id

### LINKED ISSUE #692: S3 copy APIs ignore CopySource versionId
Follow-up from #669.

S3 CopyObject and UploadPartCopy accept a versioned source object through the x-amz-copy-source header, for example /bucket/key?versionId=<id>. Kumo currently parses the whole suffix as part of the key, so it cannot copy a specific source version.

Expected behavior:

- parse versionId from x-amz-copy-source
- CopyObject should read that source object version when present
- UploadPartCopy should copy bytes from that source object version when present
- responses should expose x-amz-copy-source-version-id when a source version was used

Actual behavior:

- CopyObject and UploadPartCopy look for an object whose key includes ?versionId=...
- versioned copies fail or copy only the latest object when callers cannot express a source version

This affects SDK flows that copy from versioned buckets, including multipart copy workflows.

mizchi:
## Summary
- Fixes #692
- Related to #669
- parse versionId from x-amz-copy-source
- make CopyObject read the requested source object version
- make UploadPartCopy copy bytes from the requested source object version
- return x-amz-copy-source-version-id when a versioned source was used

## Tests
- go test ./internal/service/s3 -run TestParseCopySource|TestCopyObjectCopiesSpecifiedSourceVersion|TestUploadPartCopyCopiesSpecifiedSourceVersion -count=1
- go test ./... -count=1
- make lint
- git diff --check

### DIFF
--- a/internal/service/s3/copyobject_version_test.go
+++ b/internal/service/s3/copyobject_version_test.go
@@ -1,0 +1,109 @@
+package s3
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+func TestCopyObjectCopiesSpecifiedSourceVersion(t *testing.T) {
+	t.Parallel()
+
+	store := NewMemoryStorage()
+	svc := New(store, "")
+	ctx := context.Background()
+
+	v1, v2 := setupVersionedSourceObject(t, store)
+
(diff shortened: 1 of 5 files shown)


sivchari (approved):
LGTM. Correctly threads versionId through CopyObject and UploadPartCopy paths.

PR #695, opened 2026-05-13
fix(s3): return bucket notification config

### LINKED ISSUE #694: S3 GET bucket notification returns ListObjects response
Follow-up from #669.

Kumo accepts PUT /{bucket}?notification for EventBridgeConfiguration, but GET /{bucket}?notification is not routed to a notification handler. It falls through to the normal bucket GET path and returns a ListBucketResult instead of NotificationConfiguration.

Expected behavior:

- GET bucket notification should return a NotificationConfiguration XML document
- when EventBridge notification was enabled, it should include EventBridgeConfiguration
- missing buckets should return NoSuchBucket

Actual behavior:

- GET /{bucket}?notification returns ListBucketResult
- clients that read back bucket notification configuration, including Terraform, see drift or fail to decode the response.

mizchi:
## Summary
- Fixes #694
- Related to #669
- route GET /{bucket}?notification to a notification configuration handler
- return NotificationConfiguration XML instead of falling through to ListBucketResult
- include EventBridgeConfiguration when EventBridge notification is enabled
- return NoSuchBucket for missing buckets

## Tests
- go test ./internal/service/s3 -run TestBucketNotification_EventBridgeRoundTrip -count=1
- go test ./... -count=1
- make lint
- git diff --check

### DIFF
--- a/internal/service/s3/bucket_resources_test.go
+++ b/internal/service/s3/bucket_resources_test.go
@@ -181,6 +181,55 @@ func TestBucketLifecycle_PutGetDelete(t *testing.T) {
 	}
 }
 
+// --- BucketNotification ---------------------------------------------
+
+func TestBucketNotification_EventBridgeRoundTrip(t *testing.T) {
+	t.Parallel()
+
+	svc := New(NewMemoryStorage(), "")
+	ctx := context.Background()
+
+	store, ok := svc.storage.(*MemoryStorage)
+	if !ok {
+		t.Fatalf("storage is not *MemoryStorage")
+	}
+
+	_ = store.CreateBucket(ctx, "nb")
+
+	putReq := httptest.NewRequest(
+		http.MethodPut,
+		"/nb?notification",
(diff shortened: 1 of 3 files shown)


sivchari (changes requested):
Two issues to fix:

1. `fuzz_test.go` has `_, _, _ = parseCopySource(source)` but `parseCopySource` returns 2 values, not 3. This will cause a compile error. Likely a merge conflict resolution issue — please fix.

2. `GetBucketNotificationConfiguration` only returns `EventBridgeConfiguration` but ignores `QueueConfigurations` that are already stored via `PutBucketNotificationConfiguration`. Please include them in the response:
```go
resp.QueueConfigurations = s.storage.GetQueueConfigurations(r.Context(), bucket)
```

sivchari:
QueueConfigurations fix looks good. The fuzz_test.go issue (`_, _, _ = parseCopySource(source)` — should be two return values, not three) is still present. Please also rebase on latest main.

PR #697, opened 2026-05-13
fix(s3): version completed multipart uploads

### LINKED ISSUE #696: S3 CompleteMultipartUpload does not create object versions
Follow-up from #669.

CompleteMultipartUpload currently stores the completed object directly into the latest object map and does not apply bucket versioning rules.

Expected behavior:

- when bucket versioning is Enabled, completing a multipart upload should assign a new VersionId and make that version retrievable
- when bucket versioning is Suspended, completing a multipart upload should write the null version and replace any existing null version

Actual behavior:

- completed multipart objects always have an empty VersionId
- GetObjectVersion cannot retrieve the completed multipart object by version
- suspended buckets do not get the expected null-version behavior for completed multipart uploads

This diverges from PutObject behavior and affects SDK/Terraform flows that use multipart uploads into versioned buckets.

mizchi:
## Summary
- Fixes #696
- Related to #669
- apply the same bucket versioning rules to CompleteMultipartUpload as PutObject
- create retrievable VersionIds for completed multipart objects when versioning is enabled
- write and replace the null version when bucket versioning is suspended

## Tests
- go test ./internal/service/s3 -run TestCompleteMultipartUpload.*Version -count=1
- go test ./... -count=1
- make lint
- git diff --check

### DIFF
--- a/internal/service/s3/complete_multipart_versioning_test.go
+++ b/internal/service/s3/complete_multipart_versioning_test.go
@@ -1,0 +1,180 @@
+package s3
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+const completeMultipartVersionBody = "hello"
+
+func TestCompleteMultipartUploadCreatesVersionWhenBucketVersioningEnabled(t *testing.T) {
+	t.Parallel()
+
+	store := NewMemoryStorage()
+	ctx := context.Background()
+
+	if err := store.CreateBucket(ctx, "bucket"); err != nil {
+		t.Fatalf("CreateBucket: %v", err)
+	}
+
+	if err := store.PutBucketVersioning(ctx, "bucket", VersioningEnabled); err != nil {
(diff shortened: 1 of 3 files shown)


sivchari (changes requested):
The storage-layer fix is correct — `storeObjectVersion` properly unifies the versioning logic between `PutObject` and `CompleteMultipartUpload`.

However, the `CompleteMultipartUpload` handler does not set the `x-amz-version-id` response header when the bucket has versioning enabled. `PutObject` does this (around line 783–787), but `CompleteMultipartUpload` just calls `writeXMLResponse(w, result)` without setting the header.

AWS returns `x-amz-version-id` on `CompleteMultipartUpload` when the bucket is versioned. Without it, SDK clients will see `VersionId` as nil in the output.

Please add something like:

```go
if obj.VersionID != "" {
    w.Header().Set("x-amz-version-id", obj.VersionID)
}
```

before the `writeXMLResponse` call in the handler.

PR #700, opened 2026-05-13
fix(dynamodb): partition parallel scan segments

### LINKED ISSUE #610: DynamoDB Parallel Scan ignores Segment / TotalSegments (returns full table per segment)
## Summary

\`Scan\` with \`Segment\` / \`TotalSegments\` returns the full table for
every segment instead of partitioning items across segments. With
\`TotalSegments=8\` you get 8x duplicate items.

## Repro

\`\`\`go
client.CreateTable(...)  // PK = id String
// BatchWrite 100k items with id = strconv.Itoa(i)

var total int
for s := 0; s < 8; s++ {
    out, _ := client.Scan(ctx, &dynamodb.ScanInput{
        TableName:     aws.String(\"t\"),
        Segment:       aws.Int32(int32(s)),
        TotalSegments: aws.Int32(8),
    })
    total += len(out.Items)
}
fmt.Println(total)  // expected: 100000   actual: 800000
\`\`\`

## Where

\`internal/service/dynamodb/types.go\` defines \`Segment\` /
\`TotalSegments\` on \`ScanRequest\` but \`internal/service/dynamodb/storage.go:Scan\`
never reads them — every call iterates the whole \`td.Items\` map.

## Impact

- Any client that uses parallel scan (terraform-provider-aws, the
  official DynamoDB Streams toolkit, the \`go.dev/x/sync/errgroup\`
  scan example, etc.) will see correct items but multiplied by
  \`TotalSegments\`.
- Aggregation jobs against kumo silently overcount.
- Discovered while building a 1brc-style benchmark: setting
  \`TotalSegments=8\` produced \`scanned items: 800000\` for a 100k-item
  table.

## Suggested fix

Hash each item's primary key (e.g. fnv64 of \`serializeKey(item)\`) and
keep only items where \`hash % TotalSegments == Segment\` when those
fields are set. Match real DynamoDB behavior (segment ownership is
implementation-defined but stable per-call) by using the same hash
function on every call within a process.

## What I am proposing

I am happy to send a PR. Wanted to file the issue first to confirm
the desired hashing approach (fnv64 vs sha vs raw key bytes) so we
don't bikeshed in review.

mizchi:
## Summary
- Honor DynamoDB `Scan` requests with `Segment` and `TotalSegments`.
- Use a stable FNV-1a hash of the serialized primary key to assign each item to one segment.
- Add regression coverage that combines all segments without duplicates.

Closes #610.
Related to #669.

## Tests
- `go test ./internal/service/dynamodb -run TestScanPartitionsItemsAcrossSegments -count=1`
- `go test ./internal/service/dynamodb -count=1`
- `go test ./... -count=1`
- `make lint`
- `git diff --check`

### DIFF
--- a/internal/service/dynamodb/handlers.go
+++ b/internal/service/dynamodb/handlers.go
@@ -419,6 +419,8 @@ func (s *Service) Scan(w http.ResponseWriter, r *http.Request) {
 		req.ExpressionAttributeValues,
 		req.Limit,
 		req.ExclusiveStartKey,
+		req.Segment,
+		req.TotalSegments,
 	)
 	if err != nil {
 		var tErr *TableError
--- a/internal/service/dynamodb/parallel_scan_test.go
+++ b/internal/service/dynamodb/parallel_scan_test.go
@@ -1,0 +1,87 @@
+package dynamodb
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+func TestScanPartitionsItemsAcrossSegments(t *testing.T) {
+	t.Parallel()
+
+	store := NewMemoryStorage("http://localhost:4566")
+	svc := New(store)
+
+	if _, err := store.CreateTable(t.Context(), &CreateTableRequest{
+		TableName: "parallel-scan-test",
+		KeySchema: []KeySchemaElement{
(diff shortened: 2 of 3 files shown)


sivchari (approved):
LGTM. FNV-1a hash-based segmentation correctly satisfies the coverage + no-duplication invariant.

PR #726, opened 2026-05-18
fix(eventbridge): respect AWS_DEFAULT_REGION and resolve API destinations tolerantly

### LINKED ISSUE #725: fix(eventbridge): API destination dispatch silently dropped for cross-region target ARN
## Problem

`MemoryStorage.region` in `internal/service/eventbridge/storage.go` is hardcoded to `us-east-1` with no env override, and `resolveAPIDestination` strict-compares the full target ARN against the stored destination ARN. A target ARN whose region differs from the storage's own (e.g. an `ap-northeast-1` ARN against a `us-east-1` storage — the common case for users whose init scripts pin to their AWS production region) never resolves, so `matchAndDeliver` silently skips dispatch while `PutEvents` still returns 200 OK. EventBridge → API destination (HTTP POST) breaks entirely for those users.

## Fix

- Pick region from `AWS_DEFAULT_REGION` with `us-east-1` fallback (same convention as #520 for sfn).
- Look up API destinations by name (suffix after `:api-destination/`) so cross-region target ARNs still resolve.

tak848:
Closes #725

## Summary

- Pick region from `AWS_DEFAULT_REGION` with `us-east-1` fallback (same convention as #520 for sfn).
- Look up API destinations by name (the suffix after `:api-destination/`) so cross-region target ARNs still resolve. The name-based lookup is strictly more permissive than the previous strict-equality path, so no existing caller can be broken.

## Test plan

- [x] `TestNewMemoryStorage_RegionFromEnv` covers `AWS_DEFAULT_REGION` set / empty / fallback, plus the propagated default event bus ARN.
- [x] `TestPutEvents_APIDestinationCrossRegion` dispatches against an `httptest.Server`-backed API destination with a target ARN whose region differs from the storage's, asserting the dispatch arrives within 2s (regression guard for the silent-drop path).
- [x] `go test -race ./internal/service/eventbridge/...` passes.
- [x] `make lint` passes (0 issues).

### DIFF
--- a/internal/service/eventbridge/storage.go
+++ b/internal/service/eventbridge/storage.go
@@ -116,13 +116,18 @@ type MemoryStorage struct {
 func NewMemoryStorage(opts ...Option) *MemoryStorage {
 	ctx, cancel := context.WithCancel(context.Background())
 
+	region := os.Getenv("AWS_DEFAULT_REGION")
+	if region == "" {
+		region = "us-east-1"
+	}
+
 	s := &MemoryStorage{
 		EventBuses:      make(map[string]*EventBus),
 		Rules:           make(map[string]map[string]*Rule),
 		Targets:         make(map[string]map[string][]*Target),
 		Connections:     make(map[string]*Connection),
 		APIDestinations: make(map[string]*APIDestination),
-		region:          "us-east-1",
+		region:          region,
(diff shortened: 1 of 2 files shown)


sivchari (approved):

sivchari:
Thx!

Issue #754, opened 2026-05-26
AWS SDK for Java v2 User-Agent service detection fails for uppercase service names


OliveiraNt:
# AWS SDK for Java v2 User-Agent service detection fails for uppercase service names

## Summary

AWS SDK for Java v2 sends uppercase service names in the `api/<SERVICE>#<version>` segment of the User-Agent.

Example:

`api/SNS#2.44.x`

Kumo currently appears to use the extracted service name without normalization, which causes service detection to fail for Java SDK clients.

The issue happens because the parsed service name is `SNS` instead of `sns`.

## Reproducing

The issue can be reproduced using the AWS SDK for Java v2 against Kumo.

Example User-Agent generated by the SDK:

`aws-sdk-java/2.44.12 md/io#sync md/http#Apache ua/2.1 api/SNS#2.44.x os/Linux#6.1 lang/java#17 md/OpenJDK_64-Bit_Server_VM#17 md/vendor#OpenJDK md/en_US`

Notice that the service name is uppercase:

`api/SNS#2.44.x`

### Example test

```java
@Test
void shouldCreateSnsTopicOnLocalKumo() {
  try (SnsClient snsClient = SnsClient.builder()
      .endpointOverride(URI.create("http://127.0.0.1:4566"))
      .region(Region.US_EAST_1)
      .credentialsProvider(StaticCredentialsProvider.create(
          AwsBasicCredentials.create("test", "test")))
      .build()) {

    CreateTopicRequest request = CreateTopicRequest.builder()
        .name("meu-topico-teste")
        .build();

    CreateTopicResponse response = snsClient.createTopic(request);

    assertNotNull(response.topicArn());
  }
}
```

## Current behavior

Kumo fails to identify the AWS service because the parsed service name is `SNS` instead of `sns`.

## Expected behavior

Service detection should be case-insensitive and correctly resolve:

`SNS` → `sns`

## Proposed fix

Normalize the extracted service name before lookup/matching:

```go
service := strings.ToLower(matches[1])
```

## Additional context

Other AWS SDKs may use different casing conventions in the User-Agent, so normalizing the parsed service name improves cross-SDK compatibility.

sivchari:
This issue has been resolved.

PR #755, opened 2026-05-26
fix(server): normalize AWS service name from User-Agent

OliveiraNt:
fix issue #754


sivchari (approved):
Good catch! Thx!

sivchari (approved):
LGTM

PR #808, opened 2026-06-11
fix(dynamodb): fail closed on unparseable FilterExpression and implement IN

### LINKED ISSUE #807: DynamoDB: unparseable FilterExpression silently matches every item (IN operator unsupported — Scan/Query filters fail open)
Discovered in production e2e use (a status-cleanup job that scans for items in a set of states).

## Problem

`Scan`/`Query` with a `FilterExpression` that the expression parser cannot handle **silently matches every item** instead of returning an error. The most likely way to hit this is the `IN` operator, which the condition grammar does not implement — but any unparseable expression behaves the same way.

`evaluateFilterExpression` treats a parse error as "match":

```go
// internal/service/dynamodb/storage.go
func (m *MemoryStorage) evaluateFilterExpression(...) bool {
	result, err := evaluateCondition(item, ConditionInput{...})
	if err != nil {
		return true // ← parse error: the filter is silently ignored
	}

	return result
}
```

Notably, conditional writes already do the right thing — `PutItem`/`DeleteItem`/`UpdateItem` map the same parse error to `ValidationException: Invalid ConditionExpression: ...`. Only the read-path filter fails open. (`Query`'s `KeyConditionExpression` at the same call site swallows the error in the opposite direction: `ok, _ :=` → unparseable key condition matches nothing.)

## Impact

A filter like `#s IN (:a, :b)` returns the **entire table** with no error signal anywhere — HTTP 200, no log line. Downstream code then processes items the filter was supposed to exclude. In our case a cleanup job selected items in two states and quietly received items in every state; nothing failed until a later invariant check caught it. Comparison operators (`=`, `<`, ...) work correctly, which makes the broken case easy to miss in review.

## Expected behavior

DynamoDB supports `IN` in filter and condition expressions (up to 100 operands), and rejects invalid expressions with `ValidationException: Invalid FilterExpression: Syntax error`. See [Comparison operator and function reference](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.OperatorsAndFunctions.html).

Two changes, independently useful:

1. **Fail closed on parse errors**: an unparseable `FilterExpression` should return `ValidationException` (mirroring the existing `ConditionExpression` handling), never match-all.
2. **Implement `IN`** in the condition grammar. (`size` and `attribute_type` are also absent and currently hit the same fail-open path.)

## Reproduction (kumo 0.25.1)

```bash
docker run -d --name kumo -p 4566:4566 -e KUMO_HOST=0.0.0.0 ghcr.io/sivchari/kumo:0.25.1

E=http://localhost:4566; H='Content-Type: application/x-amz-json-1.0'
curl -s $E -H "$H" -H "X-Amz-Target: DynamoDB_20120810.CreateTable" -d '{"TableName":"probe","AttributeDefinitions":[{"AttributeName":"id","AttributeType":"S"}],"KeySchema":[{"AttributeName":"id","KeyType":"HASH"}],"BillingMode":"PAY_PER_REQUEST"}' > /dev/null
for kv in "1 pending" "2 running" "3 done"; do set -- $kv
  curl -s $E -H "$H" -H "X-Amz-Target: DynamoDB_20120810.PutItem" -d "{\"TableName\":\"probe\",\"Item\":{\"id\":{\"S\":\"$1\"},\"status\":{\"S\":\"$2\"}}}" > /dev/null
done

# control — works: returns 1 item
curl -s $E -H "$H" -H "X-Amz-Target: DynamoDB_20120810.Scan" -d '{"TableName":"probe","FilterExpression":"#s = :a","ExpressionAttributeNames":{"#s":"status"},"ExpressionAttributeValues":{":a":{"S":"pending"}}}'

# bug — IN returns all 3 items (real DynamoDB: 2)
curl -s $E -H "$H" -H "X-Amz-Target: DynamoDB_20120810.Scan" -d '{"TableName":"probe","FilterExpression":"#s IN (:a, :b)","ExpressionAttributeNames":{"#s":"status"},"ExpressionAttributeValues":{":a":{"S":"pending"},":b":{"S":"running"}}}'

# fail-open proof — garbage expression also returns all 3 items (real DynamoDB: ValidationException)
curl -s $E -H "$H" -H "X-Amz-Target: DynamoDB_20120810.Scan" -d '{"TableName":"probe","FilterExpression":"complete garbage !!!","ExpressionAttributeNames":{"#s":"status"},"ExpressionAttributeValues":{":a":{"S":"x"}}}'
```

Workaround until fixed: rewrite `IN` as an equivalent OR chain — `(#s = :a OR #s = :b)` — which the grammar handles correctly.

## Regression coverage

The follow-up should cover: `IN` with matching/non-matching operand lists on `Scan` and `Query`, and an invalid `FilterExpression` returning `ValidationException` instead of the full item set.

I'm happy to follow up with a fix PR.

wakame1367:
Closes #807

## Summary

`Scan`/`Query` treated an unparseable `FilterExpression` as a match: `evaluateFilterExpression` returned `true` on parse errors, so any unsupported syntax silently disabled filtering and returned the entire table — with HTTP 200 and no log line. The easiest way to hit this was the `IN` operator, which the condition grammar did not implement (#807).

Two changes, mirroring the issue:

- **Fail closed on parse errors.** `evaluateFilterExpression` now propagates the parse error, and `Scan`/`Query` return `ValidationException: Invalid FilterExpression: ...` — the same mapping conditional writes (`PutItem`/`DeleteItem`/`UpdateItem`) already use for `ConditionExpression`.
- **Implement `IN`.** `operand IN (operand, operand, ...)` evaluates to true when the left operand equals any list member, reusing the existing `parseArgList` and `=` comparison semantics, capped at DynamoDB's 100-operand limit. Composes with `AND`/`OR`/`NOT`/parentheses like any other comparison.

## Test plan

- [x] `make lint` (0 issues)
- [x] `go test -race ./...`
- [x] Unit: IN match/no-match, attribute-path operands, numeric operands, lowercase keyword, composition with NOT/AND/OR/parens, missing parens → error, empty list → error, 101 operands → error
- [x] Unit: Scan and Query return `ValidationException` for an unparseable filter; IN filters correctly through storage
- [x] Integration `TestDynamoDB_ScanFilterExpressionIn`: Scan with `#s IN (:a, :b)` returns Count=2/ScannedCount=3; invalid expression → `ValidationException`

### DIFF
--- a/internal/service/dynamodb/condition.go
+++ b/internal/service/dynamodb/condition.go
@@ -360,6 +360,11 @@ func parseComparison(expr string, item Item, values map[string]AttributeValue) (
 		return parseBetween(leftToken, strings.TrimSpace(rest[7:]), item, values)
 	}
 
+	// Handle IN: operand IN (operand, operand, ...)
+	if startsWithKeyword(rest, "IN") {
+		return parseIn(leftToken, strings.TrimSpace(rest[2:]), item, values)
+	}
+
 	op, afterOp, err := parseComparisonOp(rest)
 	if err != nil {
 		return false, "", err
@@ -374,6 +379,41 @@ func parseComparison(expr string, item Item, values map[string]AttributeValue) (
 	right := resolveOperand(rightToken, item, values)
 
 	result := compareAttributeValues(left, right, op)
+
+	return result, finalRest, nil
+}
+
+// maxInOperands is DynamoDB's limit on the number of operands in an IN list.
+const maxInOperands = 100
+
+// parseIn handles "operand IN (operand, operand, ...)". The result is true if
+// the left operand equals any operand in the list.
+func parseIn(leftToken, rest string, item Item, values map[string]AttributeValue) (bool, string, error) {
+	args, finalRest, err := parseArgList(rest)
+	if err != nil {
+		return false, "", fmt.Errorf("failed to parse IN operand list: %w", err)
+	}
+
+	if len(args) == 0 {
+		return false, "", fmt.Errorf("IN requires at least 1 operand")
+	}
+
+	if len(args) > maxInOperands {
+		return false, "", fmt.Errorf("IN accepts at most %d operands, got %d", maxInOperands, len(args))
+	}
+
+	val := resolveOperand(leftToken, item, values)
+
+	result := false
+
+	for _, arg := range args {
+		operand := resolveOperand(strings.TrimSpace(arg), item, values)
+		if compareAttributeValues(val, operand, "=") {
+			result = true
+
+			break
+		}
+	}
 
 	return result, finalRest, nil
 }
--- a/internal/service/dynamodb/condition_test.go
+++ b/internal/service/dynamodb/condition_test.go
@@ -3,6 +3,8 @@ package dynamodb
 import (
 	"context"
 	"errors"
+	"fmt"
+	"strings"
 	"testing"
 )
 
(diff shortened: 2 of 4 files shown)


sivchari (approved):
Thanks!

PR #855, opened 2026-07-24
fix(dynamodb): always include requested tables in BatchGetItem Responses

myohei:
## Summary

`BatchGetItem` omits the per-table key from `Responses` when none of the requested keys matched, and drops both `Responses` and `UnprocessedKeys` entirely via `omitempty` — an all-miss request returns `{}`.

Real DynamoDB always includes every requested table (as an empty list) and always serializes `UnprocessedKeys`, even when empty.

## Actual vs. expected

All-miss `BatchGetItem` (2 non-existent keys), captured against `amazon/dynamodb-local`:

```json
{"Responses":{"golden-raw-card-transactions":[]},"UnprocessedKeys":{}}
```

Same request against kumo (v0.26.0 and current `main`):

```json
{}
```

moto's reference implementation matches dynamodb-local: it unconditionally seeds `results["Responses"][table_name] = []` before iterating keys ([moto/dynamodb/responses.py](https://github.com/getmoto/moto/blob/master/moto/dynamodb/responses.py)).

## Why it matters

SDK client code commonly distinguishes "table present with zero matches" from "table missing" via key presence:

```go
avMaps, ok := res.Responses[tableName]
if !ok {
    return nil, errors.New("unexpected response from dynamodb")
}
```

This pattern works against real DynamoDB but breaks against kumo whenever a batch lookup has zero hits — e.g. an idempotency pre-check on the first import of a dataset, which is exactly how we hit this.

## Changes

- `storage.go`: build `items` as a non-nil empty slice and always set `responses[tableName]`, so zero-match tables serialize as `[]` instead of disappearing.
- `types.go`: drop `omitempty` from `BatchGetItemResponse.Responses` / `UnprocessedKeys` — DynamoDB always serializes both.
- `handlers.go`: initialize `UnprocessedKeys` to an empty map so it serializes as `{}` rather than `null`.

## Tests

- New `TestBatchGetItemResponseShape` asserts on the raw JSON body (SDK-side unmarshalling would mask omitted fields): all-miss, hit+miss mixed, and multi-table with one empty table.
- `make test` and `make lint` pass.

## Note

`BatchWriteItemResponse.UnprocessedItems` has the same class of issue (`omitempty`, real DynamoDB returns `"UnprocessedItems":{}`). Left out to keep this change focused — happy to send a follow-up PR.

### DIFF
--- a/internal/service/dynamodb/batch_get_item_response_test.go
+++ b/internal/service/dynamodb/batch_get_item_response_test.go
@@ -1,0 +1,152 @@
+package dynamodb
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+// DynamoDB always includes every requested table in Responses (as an empty
+// list when no keys matched) and always includes UnprocessedKeys, even when
+// empty. Verified against amazon/dynamodb-local:
+//
+//	{"Responses":{"table":[]},"UnprocessedKeys":{}}
+//
+// These tests assert on the raw JSON body because SDK-side unmarshalling
+// masks omitted fields.
+//
+//nolint:funlen // Exercises the response shape across all hit/miss combinations.
+func TestBatchGetItemResponseShape(t *testing.T) {
+	t.Parallel()
+
+	store := NewMemoryStorage("http://localhost:4566")
+	svc := New(store)
+
+	for _, tableName := range []string{"batch-get-hit", "batch-get-miss"} {
+		if _, err := store.CreateTable(t.Context(), &CreateTableRequest{
+			TableName: tableName,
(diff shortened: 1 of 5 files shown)


myohei:
Updated the `TestDynamoDB_BatchGetItem` golden: `"UnprocessedKeys": null` → `{}`. This is the intended behavior change — real DynamoDB always serializes `UnprocessedKeys` as `{}` when empty (verified against `amazon/dynamodb-local`).

The three `TestTerraform_*` failures (`DynamoDB_Table` / `EventBridge_Target` / `SNS_Topic`, "env=terraform tag not found") are unrelated to this change — they also fail on `main`'s latest integration run: https://github.com/sivchari/kumo/actions/runs/30092819402

PR #918, opened 2026-08-19
feat(s3): support If-Match / If-None-Match conditional writes on PutObject

shiv3:
## Motivation

Applications that use S3 conditional writes for optimistic concurrency
(e.g. a WAL bucket written by multiple writers, using `If-Match` ETag
CAS) can't be tested against kumo today: kumo silently ignores
`If-Match` / `If-None-Match` on `PutObject` and always returns 200.

Verified against kumo v0.28.1:

```console
$ curl -sS -o /dev/null -X PUT "http://localhost:4566/b/k" -d hello -w '%{http_code}\n'
200
$ curl -sS -o /dev/null -X PUT "http://localhost:4566/b/k" -H 'If-None-Match: *' -d again -w '%{http_code}\n'
200   # real S3: 412 PreconditionFailed
$ curl -sS -o /dev/null -X PUT "http://localhost:4566/b/k" -H 'If-Match: "deadbeef"' -d again -w '%{http_code}\n'
200   # real S3: 412 PreconditionFailed
```

Real S3 returns `412 PreconditionFailed` for both of these (and `409
ConditionalRequestConflict` when concurrent conditional writes collide
internally on the S3 side — out of scope here, see below).

## What's implemented

| Header | Object state | Result |
|---|---|---|
| `If-None-Match: *` | key absent (incl. current version is a delete marker) | 200, new ETag |
| `If-None-Match: *` | key exists | 412 `PreconditionFailed` |
| `If-None-Match: <etag>` (anything but `*`) | any | 501 `NotImplemented` — AWS S3 only supports `*` on writes |
| `If-Match: <etag>` | current ETag matches (quoted or unquoted) | 200, new ETag |
| `If-Match: <etag>` | current ETag differs | 412 `PreconditionFailed` |
| `If-Match: <etag>` | key absent | 404 `NoSuchKey` |
| both headers present | — | `If-Match` evaluated first (RFC 9110 §13.1 order) |
| neither header | — | unchanged (unconditional write) |

- The precondition check runs **inside the same storage lock as the
  insert** (`MemoryStorage.PutObjectIf`/`CompleteMultipartUploadIf`, see
  `checkPutCondition` in `internal/service/s3/conditional.go`), so two
  concurrent conditional `PUT`s can't both observe "no current object"
  and both succeed.
- `Storage.PutObject` and `Storage.CompleteMultipartUpload` are now thin
  wrappers around new `PutObjectIf` / `CompleteMultipartUploadIf` methods
  that take a `PutCondition{ IfMatch string; IfNoneMatchAny bool }`
  (zero value = unconditional), so the interface change is
  backward-compatible with every existing caller/test.
- **`CompleteMultipartUpload` is covered too**, with the same two
  headers/semantics as `PutObject`. A failed precondition leaves the
  in-progress upload untouched (it's deleted only on a successful
  complete), so the caller can resolve the conflict and retry — verified
  by a test that fails a conditional complete, confirms the upload
  still exists, then completes it unconditionally.
- `matchesAnyETag` now strips surrounding quotes before comparing, so an
  unquoted `If-Match`/`If-None-Match` value matches kumo's quoted stored
  ETag. This is a one-line change to a shared helper, so it also loosens
  the pre-existing `GetObject` conditional-GET and `CopyObject`
  `x-amz-copy-source-if-*` paths the same way — intentional, matches
  AWS S3's lenient comparison, and doesn't change behavior for any
  already-quoted value (all existing tests still pass unmodified).

## Out of scope

- **`409 ConditionalRequestConflict`** — this is S3's own internal
  concurrency-arbitration response when two conditional writes race
  inside S3 itself; not meaningful for a single-process emulator that
  already serializes the check-and-write under one lock.
- **`CopyObject` destination conditionals** — `CopyObject` doesn't
  accept `If-Match`/`If-None-Match` on the destination in the S3 API (it
  has its own `x-amz-copy-source-if-*` family for the *source*, which
  was already implemented and is untouched by this change).

## How tested

```console
$ make test   # 31 packages, exit 0, no failures
$ make lint   # golangci-lint, 0 issues
```

New unit tests in `internal/service/s3/conditional_test.go`:
`TestMatchesAnyETag_QuoteNormalization`, `TestParsePutCondition`,
`TestCheckPutCondition` (RFC 9110 order + delete-marker-as-absent),
`TestPutObject_ConditionalRequests` (table-driven, HTTP layer),
`TestPutObject_ConditionalRequests_DeleteMarker`,
`TestCompleteMultipartUpload_ConditionalRequest` (412 + upload survival
+ successful retry).

Also added `TestS3_PutObject_ConditionalWrites` to
`test/integration/s3_object_ops_test.go` (uses aws-sdk-go-v2's
`PutObjectInput.IfMatch`/`IfNoneMatch`). `make test-integration` needs
Docker per CONTRIBUTING.md, but the harness just points the SDK at
`KUMO_TEST_ENDPOINT`/`http://localhost:4566`, so I validated it for real
by building the binary and running it directly:

```console
$ go build -o /tmp/kumo-bin ./cmd/kumo && /tmp/kumo-bin serve &
$ go test -C test -tags=integration -run 'TestS3' ./integration/...
ok  	github.com/sivchari/kumo/test/integration	4.121s
```

(the full `TestS3*` suite, not just the new test, against a freshly
started server — all green)

## curl repro (after this change)

```console
$ curl -sS -X PUT "http://localhost:4566/b/k" -H 'If-None-Match: *' -d again
<Error><Code>PreconditionFailed</Code><Message>At least one of the preconditions you specified did not hold.</Message>...</Error>
# 412

$ curl -sS -X PUT "http://localhost:4566/b/k" -H 'If-Match: "deadbeef"' -d again
<Error><Code>PreconditionFailed</Code><Message>At least one of the preconditions you specified did not hold.</Message>...</Error>
# 412
```

### DIFF
--- a/internal/service/s3/conditional.go
+++ b/internal/service/s3/conditional.go
@@ -84,19 +84,87 @@ func evalCopySourcePreconditions(h http.Header, etag string, lastModified time.T
 // exists, which for a successful GetObject lookup is always true).
 //
 // ETag comparison is the *strong* variant per RFC 9110 §8.8.3.2 —
-// kumo doesn't emit `W/"..."` weak ETags, and AWS S3 doesn't either,
-// so quoted-string equality is sufficient.
+// kumo doesn't emit `W/"..."` weak ETags, and AWS S3 doesn't either.
+// Tokens are compared with surrounding quotes stripped so an unquoted
+// If-Match/If-None-Match value (real clients and the AWS CLI send
+// either form) still matches the quoted ETag kumo stores, matching AWS
+// S3's lenient behavior.
 func matchesAnyETag(headerValue, etag string) bool {
 	v := strings.TrimSpace(headerValue)
 	if v == "*" {
 		return true
 	}
 
+	want := strings.Trim(etag, `"`)
+
 	for _, raw := range strings.Split(v, ",") {
-		if strings.TrimSpace(raw) == etag {
+		if strings.Trim(strings.TrimSpace(raw), `"`) == want {
 			return true
 		}
 	}
 
 	return false
+}
+
+// parsePutCondition parses the If-Match / If-None-Match request headers
+// into a PutCondition for PutObject / CompleteMultipartUpload. AWS S3
+// only supports `If-None-Match: *` on writes (any other value returns
+// 501 NotImplemented instead of being evaluated), reported via the
+// second return value being false.
+func parsePutCondition(h http.Header) (PutCondition, bool) {
+	var cond PutCondition
+
+	if v := strings.TrimSpace(h.Get("If-None-Match")); v != "" {
+		if v != "*" {
+			return PutCondition{}, false
+		}
+
+		cond.IfNoneMatchAny = true
+	}
+
+	cond.IfMatch = strings.TrimSpace(h.Get("If-Match"))
+
+	return cond, true
+}
+
+// objectErrorStatus maps an *ObjectError Code produced by
+// checkPutCondition to its S3 HTTP status: PreconditionFailed is 412,
+// NoSuchKey (and anything else) is 404.
+func objectErrorStatus(code string) int {
+	if code == "PreconditionFailed" {
+		return http.StatusPreconditionFailed
+	}
+
+	return http.StatusNotFound
+}
+
+// checkPutCondition evaluates cond against bucket b's current object at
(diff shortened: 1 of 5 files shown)


sivchari (approved):
make sense to me. thx!

